PR #1, opened 2022-01-31
Setup system tests

felixclack:
We want a way to be able to run tests in a real browser environment.

[cuprite](https://github.com/rubycdp/cuprite) is a pure Ruby implementation of CDP (Chrome Developer Protocol). 
It makes it possible to run our tests in Chrome without requiring Selenium 
and the chrome webdriver.

This leads to an easier setup for developers and also a slightly faster
test suite.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,6 +52,8 @@ group :development do
 end
 
 group :test do
+  gem 'capybara', '~> 3.36'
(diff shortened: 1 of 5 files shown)


tvararu (approved):
Really clean :shipit:

PR #2, opened 2022-01-31
Add importmap

felixclack:
The importmap-rails gem is installed in the project but not setup for
use yet.

This is a tool that allows us to import JS dependencies into the project
without needing to use Webpacker.

To add a dependency in the future, run:

```bash
  bin/importmap pin react
```


tvararu (approved):
Making a mental note to make sure HTTP/2 is setup correctly, as `importmap` relies on this to get good performance.

tvararu:
Another thought: this context is ADR worthy.

felixclack:
Do you want me to add the ADR in this PR too?

tvararu:
@felixclack no need, we can write it separately, ship this :shipit:

PR #3, opened 2022-01-31
Remove turbo

felixclack:
We aren't using Turbo for anything yet. Removing it so that we can
explicitly add it in the future when we're able to define why.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -14,8 +14,6 @@ gem 'puma', '~> 5.0'
 gem 'rails', '~> 7.0.1'
 gem 'sprockets-rails'
 gem 'sqlite3', '~> 1.4'
-gem 'stimulus-rails'
-gem 'turbo-rails'
 gem 'tzinfo-data', platforms: %i[mingw mswin x64_mingw jruby]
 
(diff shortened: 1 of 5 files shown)


dankmitchell (commented):
@felixclack I think we need to remove all references to "turbo" i.e. `<%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>`

felixclack:
> @felixclack I think we need to remove all references to "turbo" i.e. `<%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>`

Good point. Removed.

dankmitchell:
> > @felixclack I think we need to remove all references to "turbo" i.e. `<%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>`
> 
> Good point. Removed.

These too @felixclack - https://github.com/DFE-Digital/find-my-trn/blob/main/package.json#L5-L6

felixclack:
Yep. Already removed those.

dankmitchell (approved):

tvararu (commented):
:+1:

PR #6, opened 2022-02-01
Add prettier for Ruby formatting

felixclack:
There is a preference within the team to use prettier for formatting
Ruby as well as our JS.

This introduces the plugin on the Ruby side (it is already enabled for
JS) and ensures it doesn't collide with the rules provided by Rubocop.

We get the best of both tools with this setup.


tvararu (approved):
<3

PR #9, opened 2022-02-02
Added outstanding ADRs

dendle:
- Decision to use Ruby on Rails 7
- Use Postgres for persistence
- Adopt GOV.UK Design system
- Monolith architecture


dendle:
Closed this PR as it clashed with an existing PR

PR #14, opened 2022-02-04
Setup deployment to PaaS

felixclack:
The PaaS service is the preferred method of deployment for this app.

The latest version of Ruby that is supported on PaaS is 3.0.3, so I had
to modify the app to use that version.

Also, the latest version of Bundler is 2.2.33, so I had to downgrade in
order for the deployment to complete successfully.


tvararu:
Creds added to GitHub actions!

Noticed some errors on https://find-my-trn-dev.london.cloudapps.digital/ (can be handled separately):

image

tvararu (approved):
Pipeline LGTM and we can always iterate on it after merge if we find issues :+1:

PR #22, opened 2022-02-10
Add a HOST environment variable to configure production

felixclack:
Rails has a feature that prevents DNS rebinding and `Host` header
attacks. It is disabled by default on production.

This change enables feature when the environment variable `HOST` is
present.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -86,4 +86,6 @@ Rails.application.configure do
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
+
+  config.hosts = [ENV['HOST']]
 end


tvararu (approved):

saliceti:
This would require configuring a HOST env variable for each environment
Also this may not work in prod where we may have 2 or more domain names
Can we disable the blocked host feature instead? It doesn't protect much. It's disabled on the other services in teacher services

felixclack:
We can pass a comma separated list of domains when there are multiple.

eg. `HOST=somedomain.com,another.com`

Not setting the `HOST` variable will disable the feature.

This gives us the option of using this if we want to.

PR #34, opened 2022-02-15
Add a confirmation page

felixclack:
When a person has completed a request for a TRN, we want to display a
confirmation screen that provides details of what to expect next.

The reference design is [here](https://find-a-lost-trn.herokuapp.com/helpdesk-request-submitted).

`TrnRequest` is intended to be a record of the request. Currently, we
have no details to store on it. This can be iterated on as we progress.

For now, we simply use the ID of the record as a reference number for
person who made the request.

I opted to store this ID in the session so that the confirmation screen
has access to it.

This also introduces a way to add a prefix to the page title from a
view.

Simply use the `content_for` method using `:page_title` as the target
and it will get injected into the page title in the <head> of the
document.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,5 +1,12 @@
 # frozen_string_literal: true
 
 class PagesController < ApplicationController
-  def start; end
+  def helpdesk_request_submitted
+    @trn_request = TrnRequest.find_by(id: session[:trn_request_id])
+    redirect_to root_url unless @trn_request
+  end
+
+  def start
+    @trn_request = TrnRequest.new
+  end
 end
--- a/app/views/pages/helpdesk_request_submitted.html.erb
+++ b/app/views/pages/helpdesk_request_submitted.html.erb
@@ -1,0 +1,26 @@
+<% content_for :page_title, 'Information Received' %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_panel(
+      title_text: 'Information received',
+      text: 'We’ll send your TRN by email or get in touch to ask for more information',
+      classes: %w[govuk-!-margin-bottom-6]
+    ) %>
+    <h2 class="govuk-heading-m">What happens next?</h2>
+    <p class='govuk-body'>
+      You’ll get an email with your TRN if we find a match. Otherwise, we’ll be in touch to ask for more information.
+    </p>
+    <p class='govuk-body'>
+      We aim to respond in 5 working days.
+    </p>
+    <h2 class="govuk-heading-m">For urgent requests, call the Teaching Regulation Agency</h2>
+    <p class="govuk-body">
+      Give the helpdesk your request number: <%= @trn_request.id %>
+    </p>
+    <ul class="govuk-list">
+      <li>Telephone: 020 7593 5394</li>
+      <li>Monday to Friday, 9am to 5pm (except public holidays)</li>
+    </ul>
+  </div>
+</div>
--- a/app/views/pages/start.html.erb
+++ b/app/views/pages/start.html.erb
@@ -43,10 +43,14 @@
       Have your National Insurance number ready, as we’ll ask for this.
     </p>
 
-    <%= govuk_start_button(
-      text: "Start now",
-      href: "#",
-      classes: 'govuk-!-margin-top-2 govuk-!-margin-bottom-9') %>
+    <%= form_with model: @trn_request do %>
+      <button class="govuk-button govuk-!-margin-top-2 govuk-!-margin-bottom-9 govuk-button--start" data-module="govuk-button">
+        Start now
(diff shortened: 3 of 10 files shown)


app/controllers/pages_controller.rb
```diff
@@ -1,5 +1,11 @@
 # frozen_string_literal: true
 
 class PagesController < ApplicationController
-  def start; end
+  def helpdesk_request_submitted
+    @trn_request = TrnRequest.find(session[:trn_request_id])
+  end
```
tvararu, line 6: Should this live in `PagesController`? My thought was that this controller would be for static pages.

app/views/pages/start.html.erb
```diff
@@ -43,10 +43,14 @@
...
-      href: "#",
-      classes: 'govuk-!-margin-top-2 govuk-!-margin-bottom-9') %>
+    <%= form_with model: @trn_request do %>
+      <button class="govuk-button govuk-!-margin-top-2 govuk-!-margin-bottom-9 govuk-button--start" data-module="govuk-button">
+        Start now
+        <svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" aria-hidden="true" focusable="false">
+          <path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z" />
+        </svg>
+      </button>
+    <% end %>
```
tvararu, line 53: Interesting, did we have to use the inline HTML content for a `govuk_start_button` because that component is technically a link?

Also we couldn't use `f.govuk_submit` I guess because that's not a start button.

We will revisit this and turn this back to a link later on, so should be fine.

spec/system/trn_requests_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe 'TRN requests', type: :system do
+  it 'completing a request' do
+    visit_home_page
+    press_start_button
+    expect_to_see_information_received_page
+    expect_page_to_include_information_received
```
tvararu, line 9: This could also be:

```suggestion
    given_i_am_on_the_home_page
    when_i_press_the_start_button
    then_i_see_the_confirmation_page
```

Last method can contain two assertions (or more, or just one, divide as you think makes sense).

There's no hard rules about this, but from [looking at how it's done in Apply](https://github.com/DFE-Digital/apply-for-teacher-training/blob/bdbcb2dec64190eb27674783654368e53a2f7946/spec/system/candidate_interface/feedback/candidate_arrives_from_find_to_provide_feedback_spec.rb#L7-L28), it works better when the method names loosely follow a "given (situation) / when (action) / then (outcome)" structure. It makes the method names read like documentation.

tvararu (approved):

app/views/pages/helpdesk_request_submitted.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    </p>
+    <p class='govuk-body'>
+      We aim to respond in 5 working days.
+    </p>
+
+    <h2 class="govuk-heading-m">For urgent requests, call the Teaching Regulation Agency</h2>
+
+    <p class="govuk-body">
+      Give the helpdesk your request number: <%= @trn_request&.id %>
+    </p>
```
fofr, line 26: Does this need a check to make sure the id is present, and if not don't show the whole paragraph?

app/views/pages/start.html.erb
fofr: Why does this need to be a form here?

app/controllers/pages_controller.rb
felixclack: I wasn't sure if we would need a show action for the `TrnRequest` resource, so figured it would be easiest for now to use `PagesController` to get the desired URL and defer the decision to a later point when we know more.

app/views/pages/start.html.erb
tvararu: @fofr it's temporary. Because we're building the application backwards, we have to take users directly to the confirmation page. We need to create an entry in the database to show users. The idiomatic way to do this is using a POST request. It's possible to do it via a GET, but this has unintended consequences.

app/views/pages/helpdesk_request_submitted.html.erb
tvararu: I think rather there shouldn't be any optional chaining here; the controller should redirect to the start page if a user tries to reach this page without a `@trn_request`... The page should blow up if `id` isn't defined.

WDYT @felixclack?

felixclack: Yeah, makes sense to me.

spec/system/trn_requests_spec.rb
felixclack: 👍

tvararu (approved):

app/views/pages/helpdesk_request_submitted.html.erb
fofr: I thought sometimes the Zendesk ticket number sometimes wouldn't be available, and that's ok?

PR #36, opened 2022-02-15
Production ready deployment pipeline

merrilmathew:
### Context

Add production ready deployment pipeline with terraform support. This PR will deploy the app and backing services using Teacher services standards and follows other TRA deployments.

### Trello card

https://trello.com/c/Uxm2n5Zh

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,15 +5,10 @@ on:
     branches:
       - '**'
 
-env:
-  CONTAINER_REGISTRY: ghcr.io
-
 jobs:
   rspec:
     name: Build & Test
-    runs-on: ubuntu-20.04
-    outputs:
-      docker_image: ${{ steps.image.outputs.tag }}
+    runs-on: ubuntu-latest
 
     services:
       postgres:
@@ -48,22 +43,3 @@ jobs:
 
       - name: Run tests
         run: bundle exec rspec --format documentation
-
-      - name: Docker image tag
-        id: image
-        run: |
-          echo ::set-output name=tag::$CONTAINER_REGISTRY/$(echo $GITHUB_REPOSITORY | tr '[:upper:]' '[:lower:]'):$GITHUB_SHA
-
-      - name: Login to GitHub Container Registry
-        uses: docker/login-action@v1
-        with:
-          registry: ${{ env.CONTAINER_REGISTRY }}
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Docker build & push
-        uses: docker/build-push-action@v2
-        with:
-          context: .
-          push: true
-          tags: ${{ steps.image.outputs.tag }}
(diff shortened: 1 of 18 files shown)


merrilmathew:
[Production-ready deployment pipeline and DevOps](https://trello.com/c/Uxm2n5Zh/121-production-ready-deployment-pipeline-and-devops)

Dockerfile
```diff
@@ -2,18 +2,23 @@ FROM ruby:3.0.3-alpine
...
+
+COPY package.json /myapp/
+COPY yarn.lock /myapp/
+RUN yarn install --frozen-lockfile
+
 ADD . /myapp
+ENV RAILS_ENV=production \
+    RAILS_SERVE_STATIC_FILES=yes \
+    LANG=en_GB.UTF-8 \
+    SECRET_KEY_BASE=TestKey
```
tvararu, line 19: Here we're generating the docker image using this temporary key, but when we deploy later, are we using a different `SECRET_KEY_BASE` in the application environment?

---

[More background on what SECRET_KEY_BASE does](https://stackoverflow.com/questions/25426940/what-is-the-use-of-secret-key-base-in-rails-4/25427302#25427302).

TL;DR it can be anything, including a randomly generated GUID that we don't have to know / retrieve ever. It shouldn't live in an encrypted storage like KeyVault because we should only write it once and then forget about it.

When it gets replaced, all session cookies become invalid, so it shouldn't be changed between routine deployment (it should be persistent).

It should also be different between all envs.

merrilmathew: used terraform random_uuid to pass in env variable on deployment

terraform/app.tf
```diff
@@ -0,0 +1,44 @@
+resource "random_uuid" "id" {
+}
+locals {
+  secret_key_base = {
+    SECRET_KEY_BASE = "${random_uuid.id.result}"
```
tvararu, line 5: Just to check: this won't change SECRET_KEY_BASE between deployments right? Just once on the initial deploy.

tvararu (approved):
Can be merged as-is and then we can write another PR to fix the `test.yml` and split it up so that tests still run on branches/PRs.

terraform/app.tf
merrilmathew: it will change each time it deploys. Do you want one constant value for each environment?

tvararu: Yeah. We need a constant value, because otherwise cookies will stop working between each deploy, and log users out.

.github/workflows/test.yml
```diff
@@ -1,17 +1,35 @@
...
+    inputs:
+      environment:
+        description: 'Deploy environment ( dev, preprod or production )'
+        required: true
+        default: dev
+        type: environment
   push:
     branches:
-      - '**'
+      - main
```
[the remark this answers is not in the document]
merrilmathew: ok. Let me try to fix that.

tvararu (approved):

PR #40, opened 2022-02-17
Snags

tvararu:
### Context

A few snags I've found in our setup/documentation.

See the list of commits.

### Guidance to review

Reviewing commit by commit is easiest.

### Checklist

- [x] ~~Attach to Trello card~~ N/A
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,26 +1,37 @@
 FROM ruby:3.0.3-alpine
 
-RUN apk add gcc git libc6-compat libc-dev make nodejs postgresql13-dev \
-    sqlite-dev tzdata yarn
+RUN apk -U upgrade && \
+    apk add --update --no-cache gcc git libc6-compat libc-dev make nodejs \
+    postgresql13-dev tzdata yarn
 
-WORKDIR /app
-
-COPY Gemfile Gemfile.lock ./
-RUN gem update --system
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,25 +1,37 @@
...
+    rm -rf /usr/local/bundle/cache
+
+COPY package.json yarn.lock ./
+
+RUN yarn install --frozen-lockfile --check-files
+
+COPY . .
+
+RUN yarn build && yarn build:css && \
+    bundle exec rails assets:precompile && \
```
felixclack, line 33: I think the jsbundling and cssbundling libraries handles this for us.

https://github.com/rails/jsbundling-rails/blob/main/lib/tasks/jsbundling/build.rake
https://github.com/rails/cssbundling-rails/blob/main/lib/tasks/cssbundling/build.rake

```suggestion
RUN bundle exec rails assets:precompile && \
```

felixclack (approved):

Dockerfile
tvararu: Interesting. I don't think it's working like that currently. I'll dig into it.

tvararu: I'll open a follow-up PR to add `jdbundling`.

PR #41, opened 2022-02-17
Add the email page

felixclack:
We want to allow people to enter an email address as part of the TRN
request.

This adds a new page for collecting the email address.

We're able to change the start button on the start page back to a link
as it no longer needs to POST to create the `TrnRequest`.

I'm not sure how long the `TrnRequest` object will continue as we have
talked about using a `Teacher` object for the long term. I opted to
leave the models as they are for now, with the assumption that it is
relatively trivial to change a model name in the future.

The `edit` action for the `EmailController` also takes on the
responsibility of creating the `TrnRequest` object. I expect this to move in
the future to whatever screen precedes it.

`ValidForNotifyValidator` is a direct copy of the validator originally
written in
[apply-for-teacher-training](https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/app/validators/valid_for_notify_validator.rb/).

The assumption is that emails will be sent using Notify, so the
validations should line up with the expectations of that system.

I excluded the system specs from the `Rspec/ExampleLength` cop because
the default of 5 lines is restrictive when writing scenarios that have
multiple steps to test the same behaviour across multiple screens.

This seemed better to me than the alternative of writing a separate
scenario for each page, eg. testing the back link.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/views/email/edit.html.erb
```diff
@@ -0,0 +1,17 @@
+<% content_for :page_title, 'Your email address' %>
+<%= content_for :back_link_url, controller.request.env['HTTP_REFERER'].include?('check-answers') ? check_answers_path : start_path %>
```
tvararu, line 2: Yeah... I've been down this path...

This won't work if the user refreshes the webpage; the referrer will get lost. And the back path will reset to the start page, which isn't the page they came from...

There's an alternative solution using `history.back` from the browser, but that doesn't work without JavaScript, and can also be fooled sometimes.

I don't think it's avoidable, but we need to somehow keep track of a bit of state which can be used to answer "Has this user reached the check your answers page? If yes, take them back there. If not, take them to the previous step in the form."

So we'd probably need a boolean that gets set to `true` when we reach the check your answers page, unless we can infer this state based on other information on the `TrnRequest` object.

cc @fofr I know you've probably spent a while dealing with / thinking about this problem.

app/views/email/edit.html.erb
```diff
@@ -0,0 +1,17 @@
+<% content_for :page_title, 'Your email address' %>
```
tvararu, line 1: [Design system says](https://design-system.service.gov.uk/patterns/validation/#how-to-tell-the-user-about-validation-errors):

> If the user’s answers fail validation:
> - add ‘Error: ’ to the beginning of the page <title> so screen readers read it out as soon as possible

So `Error:` should be prefixed here if there are errors... might be worth doing it in the layout or somehow abstracting it so we don't have to do it on every form.

tvararu (commented):
Solid work :+1: just some nits. `Error:` and the back button are the only blocking things.

app/views/email/edit.html.erb
felixclack: 👍

spec/models/trn_request_spec.rb
```diff
@@ -23,4 +23,10 @@
       end
     end
   end
+
+  it do
+    expect(trn_request).to validate_presence_of(:email).with_message(
+      'Enter an email address in the correct format, like name@example.com',
```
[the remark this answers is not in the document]
tvararu: Interesting. What do you think, is the cop worth it? I'm fine either way.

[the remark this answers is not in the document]
felixclack: I'm happy with the cop. It adds friction for changing copy but I think that's a good thing. Prevents accidental changes from going under the radar.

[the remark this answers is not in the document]
tvararu: 👍

app/views/email/edit.html.erb
felixclack: How about we use the referrer when it exists and when it is absent, we check the state of the `TrnRequest#email` field and use its presence as an indicator that the check answers screen should be the target of the back link?

app/views/email/edit.html.erb
```diff
@@ -7,6 +7,7 @@
       <%= f.govuk_error_summary %>
       <%= f.govuk_email_field(
             :email, 
+            autofocus: true,
```
tvararu, line 10: Not sure we should use `autofocus`, the user's focus should just follow the page naturally.

app/views/email/edit.html.erb
tvararu: Hmm, I think that could work. 👍

app/views/email/edit.html.erb
felixclack: Ok. 👍

tvararu (approved):
Solid testing 👍

app/helpers/email_helper.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+module EmailHelper
+  def back_link_url(trn_request)
```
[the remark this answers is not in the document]
felixclack: Yeah, I thought the abstraction might be clearer when we have multiple places to call this from.

PR #47, opened 2022-02-22
Add an ITT Provider page

felixclack:
As part of requesting a TRN, we want the requestor to have the option to
provide details about where they trained. This is an optional answer.

The [reference design](https://find-a-lost-trn.herokuapp.com/itt-provider) demonstrates the usage of
the GOVUK radio buttons with an optional additional input.

It seems this particular component throws a peculiar issue in Capybara.
The parent radio buttons aren't visible to Capybara. Possibly this is
unique to Cuprite/Ferrum that is providing the connection to the browser
in tests.

The solution we landed on was to use `visible: false` to permit Capybara
to interact with elements it classes as invisible.

Ordinarily we wouldn't permit interacting with invisible elements.
However, in this case no accessibility or visual issues were uncovered
during the investigation. As a result we believe that this particular
exception to the rule is safe to proceed with.

Other assumptions made in this change:
* we're storing the ITT provider as plain text for a first iteration but
  this could change in the future to an API powered lookup of providers.
* it seems likely that TrnRequest could make use of a state machine to
  allow validations to be more declarative based on state. I resisted
  making that change currently to allow more context to emerge as future
  steps get added.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


app/models/trn_request.rb
```diff
@@ -1,10 +1,16 @@
 # frozen_string_literal: true
 class TrnRequest < ApplicationRecord
-  validates :email, valid_for_notify: true
+  validates :email, valid_for_notify: true, if: %i[itt_provider_answered? email]
```
tvararu, line 3: > It seems likely that TrnRequest could make use of a state machine to allow validations to be more declarative based on state.

I see what you mean here.

Another option is creating `FormObject`s that are temporarily instantiated for a particular step in the form. They are instantiated from a `TrnRequest` and saving one writes to a `TrnRequest`, after which it's discarded. The form objects have validations, the `TrnRequest` doesn't.

Another option is to use validation scopes on the `TrnRequest`, and use `valid?(:email)` where appropriate. This is a bit more brittle in my experience.

tvararu (approved):
:star:

app/models/trn_request.rb
felixclack: Yeah, some good ideas there. I think seeing more of the whole flow will make this decision clearer. I'll keep waiting.

PR #49, opened 2022-02-23
Add NI number to the request flow

felixclack:
The designs for the TRN request flow have steps to collect an NI number.

This is an optional step and split into 2 screens. The first to
determine if the person knows their NI number and the second to allow
them to enter it.

I tried a different approach to validating the TrnRequest. Rather than
adding a conditional validation to the `TrnRequest` class, I created a
form object, `NiNumber` that gets used to specifically add the
`ni_number` value and makes it easier to conditionally validate that
value.

It means the object representing the record doesn't need to know about
rules related to the flow.

This makes this particular object easier to understand. It comes at the
cost of spreading validation rules across multiple other objects.

At this stage, I think the trade-off is appropriate.

Assumptions:
* we validate the format of the number but have no way of checking if it
  is a real national insurance number.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,14 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
+  def back_link_url(trn_request)
+    referer = controller.request.env['HTTP_REFERER']
+    return check_answers_path if referer&.include?('check-answers') || trn_request&.email?
+
+    start_path
+  end
+
+  def pretty_ni_number(ni_number)
+    ni_number.scan(/..?/).join(' ').upcase
+  end
 end
--- a/app/models/ni_number.rb
+++ b/app/models/ni_number.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+class NiNumber
+  include ActiveModel::Model
+
+  attr_accessor :trn_request_id
+
+  validates :ni_number, format: { with: /\A[a-z]{2}[0-9]{6}[a-d]{1}\Z/i }, presence: true
+
+  delegate :email?, :ni_number, to: :trn_request
+
+  def ni_number=(value)
+    trn_request.ni_number = value&.gsub(/\s/, '')
+  end
+
+  def update(params = {})
+    self.ni_number = params[:ni_number]
+    return false if invalid?
+
+    trn_request.update(params)
+  end
+
+  private
+
+  def trn_request
+    @trn_request ||= TrnRequest.find_by(id: trn_request_id) || TrnRequest.new
+  end
+end
--- a/spec/models/ni_number_spec.rb
+++ b/spec/models/ni_number_spec.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe NiNumber, type: :model do
+  subject(:ni_number_form) { described_class.new }
+
+  specify do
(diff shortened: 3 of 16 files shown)


app/models/ni_number.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+class NiNumber
+  include ActiveModel::Model
+
+  attr_accessor :trn_request_id
+
+  validates :ni_number,
+            format: {
+              with: /\A[a-z]{2}[0-9]{6}[a-d]{1}\Z/i,
+              message: 'Enter a National Insurance number in the correct format',
```
tvararu, line 10: Maybe this? (untested code, maybe needs `activerecord.` prefix)

```suggestion
              message: I18n.t('errors.models.ni_number.attributes.ni_number.blank')
```

I think not using `I18n.t` in the tests makes sense, but what about model valdiations?

felixclack: I think I can simply remove this line from the model and it will use the locales file.

felixclack: Good call.

app/models/ni_number.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+class NiNumber
+  include ActiveModel::Model
+
+  attr_accessor :trn_request_id
+
+  validates :ni_number,
+            format: {
+              with: /\A[a-z]{2}[0-9]{6}[a-d]{1}\Z/i,
```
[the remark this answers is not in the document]
fofr: Should we also playback what's entered in the standard format on the Check answers page? (eg QQ 12 34 56 C)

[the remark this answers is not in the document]
tvararu: @fofr do you mean in the error message?

```
Enter a National Insurance number in the correct format, for example ‘QQ 12 34 56 C’
```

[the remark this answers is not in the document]
felixclack: I like it. Done.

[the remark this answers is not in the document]
felixclack: Screen Shot 2022-02-28 at 11 43 27 am

app/helpers/application_helper.rb
```diff
@@ -1,4 +1,14 @@
...
 module ApplicationHelper
+  def back_link_url(trn_request)
+    referer = controller.request.env['HTTP_REFERER']
+    return check_answers_path if referer&.include?('check-answers') || trn_request&.email?
+
+    start_path
+  end
+
+  def pretty_ni_number(ni_number)
+    ni_number.scan(/..?/).join(' ').upcase
```
tvararu, line 12: Nice!

spec/models/ni_number_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+      ' QQ 12 34 56 C',
+      'QQ12 34 56 C ',
+      'QQ 1234 56 C',
+      'QQ 12 3456 C',
+      'QQ 12 34 56C',
+      'QQ1234 56 C',
+      'QQ 123456 C',
+      'QQ 12 3456C',
+      'QQ123456C',
+      'QQ 1 2 3 4 5 6 C',
```
tvararu, line 70: Could you add one lowercase case too?

```suggestion
      'QQ 1 2 3 4 5 6 C',
      'qq123456c',
```

tvararu (approved):
:shipit:

PR #55, opened 2022-03-01
Run terraform linter in CI

tvararu:
We previously would only run the Terraform formatter on the main branch, but this can cause deploys to fail.

Also remove linting from the actual prod deploy script.


.github/workflows/lint.yml
```diff
@@ -38,3 +38,25 @@ jobs:
...
+
+      - name: Check formatting
+        run: terraform fmt -check
+        working-directory: terraform
+
+      - name: Validate
+        run: |
+          terraform init -backend=false
+          terraform validate -no-color
+        working-directory: terraform
```
tvararu, line 62: What does this do exactly @merrilmathew and should we keep it in the linting script? Should we have it in the deploy script as well?

merrilmathew: This simply validates the terraform files and configuration

merrilmathew (approved):

PR #62, opened 2022-03-02
Update wording of NI number error

tvararu:
As suggested by @fofr. :shipit:

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -40,5 +40,5 @@ en:
         ni_number:
           attributes:
             ni_number:
-              blank: Enter a National Insurance number in the correct format
-              invalid: Enter a National Insurance number in the correct format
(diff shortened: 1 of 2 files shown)


felixclack (approved):

config/locales/en.yml
```diff
@@ -40,5 +40,5 @@ en:
         ni_number:
           attributes:
             ni_number:
-              blank: Enter a National Insurance number in the correct format
-              invalid: Enter a National Insurance number in the correct format
+              blank: Enter a National Insurance number
+              invalid: Enter a National Insurance number
```
fofr, line 44: Only blank should use the shorter one.

fofr:
```
blank: Enter a National Insurance number
invalid: Enter a National Insurance number in the correct format
```

PR #65, opened 2022-03-03
Validate for presence of NI number first

tvararu:
Without this, the blank error never shows up in practice.


felixclack (approved):
TIL. I didn't realise the validations ran in order of their definition here.

PR #66, opened 2022-03-03
Rework provider validation

tvararu:
### Context

Fix the provider validations.

Slightly refactor/reorganise the main system spec.

Add `binding.pry` for easier debugging.

The new `IttProviderForm` model introduced in this PR is probably an idiomatic example of how we should do the other form objects; if we agree on this implementation, we should follow-up by going back and refactoring `NiNumber` and Email validation to use a similar model.

### Guidance to review

Review commit by commit.

### Checklist

https://trello.com/c/9jqIgjZq

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/itt_providers_controller.rb
+++ b/app/controllers/itt_providers_controller.rb
@@ -1,11 +1,12 @@
 # frozen_string_literal: true
 class IttProvidersController < ApplicationController
   def edit
-    trn_request
+    @itt_provider_form = IttProviderForm.new(trn_request: trn_request).assign_form_values
   end
 
   def update
-    if trn_request.update(itt_provider_params)
+    @itt_provider_form = IttProviderForm.new(itt_provider_params)
+    if @itt_provider_form.save
       redirect_to trn_request.email.blank? ? email_url : check_answers_url
     else
       render :edit
@@ -15,7 +16,10 @@ class IttProvidersController < ApplicationController
   private
 
   def itt_provider_params
-    params.require(:trn_request).permit(:itt_provider_enrolled, :itt_provider_name)
+    params
+      .require(:itt_provider_form)
+      .permit(:itt_provider_enrolled, :itt_provider_name)
+      .merge(trn_request: trn_request)
   end
 
   def trn_request
--- a/app/models/itt_provider_form.rb
+++ b/app/models/itt_provider_form.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+class IttProviderForm
+  include ActiveModel::Model
+
+  attr_accessor :trn_request, :itt_provider_enrolled, :itt_provider_name
+
+  validates :itt_provider_enrolled, inclusion: { in: %w[true false] }
+  validates :itt_provider_name, presence: true, length: { maximum: 255 }, if: -> { itt_provider_enrolled == 'true' }
+
+  def save
+    return false if invalid?
+
+    trn_request.update(itt_provider_enrolled: itt_provider_enrolled, itt_provider_name: itt_provider_name)
+  end
+
+  def assign_form_values
+    self.itt_provider_enrolled = trn_request.itt_provider_enrolled
+    self.itt_provider_name = trn_request.itt_provider_name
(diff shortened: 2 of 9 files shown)


app/models/itt_provider_form.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  attr_accessor :trn_request, :itt_provider_enrolled, :itt_provider_name
+
+  validates :itt_provider_enrolled, inclusion: { in: %w[yes no] }
+  validates :itt_provider_name, presence: true, length: { maximum: 255 }, if: -> { itt_provider_enrolled == 'yes' }
+
+  def save
+    return false if invalid?
+
+    trn_request.update!(itt_provider_enrolled: itt_provider_enrolled, itt_provider_name: itt_provider_name)
```
felixclack, line 13: We should probably use `update` without the bang so as to keep with the expected behaviour of `save`.

spec/models/itt_provider_form_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  describe '#save' do
+    context 'when itt_provider_enrolled is missing' do
+      it 'returns false and has errors' do
+        form = described_class.new
+
+        expect(form.save).to eq false
+        expect(form.errors.full_messages).to eq [
+             'Itt provider enrolled Tell us if you’ve been enrolled in initial teacher training',
+           ]
+      end
```
felixclack, line 14: I generally subscribe to the single assertion per test approach. This would be easy to change to align with that.

We can also use the shoulda_matchers to one line these.

```suggestion
      subect(:form) { described_class.new }
      
      specify do
        expect(form)
          .to validate_presence_of(:itt_provider_enrolled)
          .with_message('Itt provider enrolled Tell us if you’ve been enrolled in initial teacher training')
      end
```

felixclack (approved):
A few thoughts/questions but nothing blocking.

app/controllers/itt_providers_controller.rb
```diff
@@ -16,7 +16,10 @@ def update
   private
 
   def itt_provider_params
-    params.require(:trn_request).permit(:itt_provider_enrolled, :itt_provider_name)
+    params
+      .require(:itt_provider_form)
+      .permit(:itt_provider_enrolled, :itt_provider_name)
+      .merge(trn_request: trn_request)
```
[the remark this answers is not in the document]
tvararu: Good question... I think fetching feels like it should happen in the controller. We have to do it anyway, this way we avoid performing multiple fetches in the same request lifecycle. Does that make sense?

spec/models/itt_provider_form_spec.rb
tvararu: That code isn't doing the same thing though, is it? It's checking for validations but it's not checking if the save method is running those validations. I was thinking that testing it like that would be longer than doing it "in line" as I did it.

I don't have a dog in the multiple assertions race; happy either way, but generally prefer to rely less on `subject` and `before` hooks and other things as I think it makes some tests harder to figure out than "just read everything that happens in the `it` block". That's personal preference and I'm happy to align one way or another.

felixclack: This is not a blocker for me, so consider the below as me articulating how I think about tests rather than a request for a change. I'm happy to go with your approach if you prefer this after reading my thoughts.

I take your point that these tests aren't the same. However, I think this speaks to a more subtle difference here.

My preference for the single assertion is related to the single responsibility principle that I find to be very useful across all areas of software development.

When a test/method/controller/PR is responsible for one thing, I find it much easier to reason about.

Sure, we can lose the benefit of this simplicity when we overload `subject` and `before` blocks. I think this is more incorrect usage of those tools rather than them being bad.

I think the specifics of the validations are a separate concern from testing that `save` runs validations.

I'd write individual tests for each validation and then a separate describe block for `#save`. With assertions for the return value and also that the `errors` method populates.

felixclack (approved):

PR #71, opened 2022-03-07
Validate the NI number correctly

felixclack:
The validations for the NI number fields aren't triggering correctly in
the UI.

The 2 issues are that the error message for the validation is the Rails
default rather than the GOVUK style, which makes identifying the issue
hard. The GOVUK component doesn't render the field name in the errors,
so it relies on the error message including it.

Also, the ITT provider enrolled validation is firing at the wrong time.
This needs to be extracted to its own form object (which will happen in
another PR).

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/models/ni_number.rb
+++ b/app/models/ni_number.rb
@@ -16,7 +16,7 @@ class NiNumber
     self.ni_number = params[:ni_number]
     return false if invalid?
 
-    trn_request.update(params)
+    trn_request.save!
   end
 
   private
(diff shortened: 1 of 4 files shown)


tvararu (approved):
Works great locally. :+1: 

The only thing missing is that clicking the error should focus the `Yes` input field for the user. Can that be rolled into this other card? https://trello.com/c/qX7NyuAX

felixclack:
When I click the error link locally in Chrome, I get the input highlighted.

eg.
Screen Shot 2022-03-07 at 2 15 55 pm

Do you get something different?

tvararu:
@felixclack ahhh, I figured out what the issue is, it doesn't work without JavaScript! I never noticed that this behaviour requires JS in the design system. Nevermind my comment then.

PR #75, opened 2022-03-08
Add a form object for has_ni_number

felixclack:
To keep the underlying `TrnRequest` object simple and independent of
validations for specific questions, we favour form objects that can
focus on validating the values for specific questions.

This allows us to validate simply without having to factor in logic for
the state of other questions.

`has_ni_number` is currently validated on `TrnRequest` but this makes it
hard to add new questions/fields to the flow.

Extracting it to a form object means we can insert new
fields/validations in the flow without having to ensure the
`has_ni_number` question has been answered first.

It looks like we have a common format to our form objects emerging.
There may be a case for abstracting the common parts into a module we
can include in all of them.

I'm happy to leave that to a future change.

I also took the opportunity to move the files to `app/forms` to help
distinguish these form objects as different from data models.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


tvararu (approved):
Works well locally :+1:

PR #82, opened 2022-03-10
Add a name question

felixclack:
We want to collect a person's name to help match their TRN.

Previous names can be helpful in the matching process too but are
optional.

Assumptions:
* no validations required for the previous name fields

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/forms/name_form.rb
+++ b/app/forms/name_form.rb
@@ -16,6 +16,10 @@ class NameForm
     @first_name ||= trn_request&.first_name
   end
 
+  def name_changed
+    previous_first_name.present? || previous_last_name.present?
+  end
+
   def previous_first_name
     @previous_first_name ||= trn_request&.previous_first_name
   end
--- a/app/forms/name_form.rb
+++ b/app/forms/name_form.rb
@@ -3,7 +3,7 @@ class NameForm
   include ActiveModel::Model
 
   attr_accessor :trn_request
-  attr_writer :first_name, :previous_first_name, :previous_last_name, :last_name
+  attr_writer :first_name, :previous_first_name, :previous_last_name, :last_name, :name_changed
 
   validates :first_name, presence: true, length: { maximum: 255 }
   validates :last_name, presence: true, length: { maximum: 255 }
@@ -17,14 +17,20 @@ class NameForm
   end
 
   def name_changed
-    previous_first_name.present? || previous_last_name.present?
+    return @name_changed unless @name_changed.nil?
+
+    @name_changed ||= previous_first_name.present? || previous_last_name.present? || trn_request.previous_name?
   end
 
   def previous_first_name
+    return nil if !@name_changed.nil? && !@name_changed
+
     @previous_first_name ||= trn_request&.previous_first_name
   end
 
   def previous_last_name
+    return nil if !@name_changed.nil? && !@name_changed
+
     @previous_last_name ||= trn_request&.previous_last_name
   end
 
--- a/app/forms/name_form.rb
+++ b/app/forms/name_form.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+class NameForm
+  include ActiveModel::Model
+
+  attr_accessor :trn_request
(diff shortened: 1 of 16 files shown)


app/views/trn_requests/show.html.erb
```diff
@@ -7,6 +7,11 @@
     <p class="govuk-body">We will send your TRN to <%= @trn_request.email %> if we find one matching your details.</p>
 
     <% rows = [
+        { 
+          key: { text: 'Your Name' },
```
tvararu, line 11:
```suggestion
          key: { text: 'Name' }, 
```

app/views/trn_requests/show.html.erb
```diff
@@ -7,6 +7,11 @@
     <p class="govuk-body">We will send your TRN to <%= @trn_request.email %> if we find one matching your details.</p>
 
     <% rows = [
+        { 
+          key: { text: 'Your Name' }, 
+          value: { text: @trn_request.name }, 
+          actions: [{ href: name_path, visually_hidden_text: 'name' }]
+        },
```
tvararu, line 14: Could we also play back `Previous name` here as a separate section?

app/forms/name_form.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+class NameForm
+  include ActiveModel::Model
+
+  attr_accessor :trn_request
+  attr_writer :first_name, :previous_first_name, :previous_last_name, :last_name
+
+  validates :first_name, presence: true
+  validates :last_name, presence: true
```
tvararu, line 9: Could we add a max length validation of 255 (+ copy the message from another form) to all 4 fields here?

app/views/name/edit.html.erb
```diff
@@ -0,0 +1,30 @@
...
+      <%= f.govuk_text_field(:first_name,
+            label: { size: 'm', text: 'First name' },
+            required: true
+          ) %>
+      <%= f.govuk_text_field(:last_name,
+            label: { size: 'm', text: 'Last name' },
+            required: true
+          ) %>
+
+      <%= f.govuk_check_boxes_fieldset :name_changed, legend: nil do %>
```
tvararu, line 21: Found a small bug locally; when clicking the `Change name` link on the check answers page when I've entered a previous name, the checkbox renders unticked, so the previous names are hidden on initial page load.

tvararu (commented):
Looking really good! Ready to approve after the suggested tweaks.

app/forms/name_form.rb
tvararu: @fofr I read your design history post at https://tra-digital-design-history.herokuapp.com/find-a-lost-trn/names/

One suggested example is that the name "Psy", with no first/last name, should actually be a valid "name."

Is that the case in the MVP? Or one to address later if we decide to move to a combined single `Name` field?

fofr: Kate Lee had some stats on how often this occurred – while there are some records with no first or last names (more commonly first than last), it represents <0.01% in under 65s. So not in MVP.

If we track the validation messages we can see how frequently people attempt to enter only one name.

felixclack: 👍

app/views/name/edit.html.erb
felixclack: Fixed. 👍

app/views/trn_requests/show.html.erb
felixclack: Done.

Screen Shot 2022-03-11 at 8 59 04 am

tvararu: Nice, could they be combined into just `Previous name`?

felixclack: Sure. So, if only one previous name field is present, combine it with the other existing name to get a full name?

eg. Joan Smith enters a previous last name of Doe, so we display Joan Doe.

fofr: I like this approach

tvararu (approved):

PR #87, opened 2022-03-14
Move email validations to a form object

felixclack:
The email validations for a TrnRequest are currently located on the
model, which requires a state check to see if we can run the validation
or not.

Elsewhere in the request flow, we have opted for a pattern where a form
object defines the validations, which decouples these validations from
the current state of the object. This makes it easier to reason about
those validations and also easier to change the request flow.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


tvararu (approved):
:100:

PR #102, opened 2022-03-16
Ensure the back link works in all scenarios

felixclack:
There are cases where the back link doesn't behave as a user might
expect.

eg. when a form renders a validation message like a field is blank, then
the back link currently goes back to the same page but without the
error.

This is strange from a user's point of view.

Instead, we want the back link to return to the last unique page the
user was on.

I've opted to use Javascript to approximate this. The implicit
assumption here is that when the referer for a request is the same, we
assume that we need to go back 2 pages in the browser history.

This doesn't work if the form has failed validation multiple times.

For this iteration, I decided that this will handle more of the error
cases than we currently are, without having to use a session variable to
store the page of origin.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,12 +1,10 @@
 # frozen_string_literal: true
 
 module ApplicationHelper
-  def back_link_url(trn_request)
-    referer = controller.request.env['HTTP_REFERER']
-    return check_answers_path if referer&.include?('check-answers') || trn_request&.email?
-    return referer if referer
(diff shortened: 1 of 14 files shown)


tvararu (approved):
Good interim solution, but yes we should still address the situation where there are multiple validation errors in a row.

tvararu:
Closing in favor of https://github.com/DFE-Digital/find-a-lost-trn/pull/104

PR #105, opened 2022-03-21
Find a TRN using the DQT API

felixclack:
There is an API we can use to find a TRN and confirm the details of a
teacher.

### Context

Include relevant motivation and context.

Include a summary of the change.

### Guidance to review

Include any useful information needed to review this change.
Include any dependencies that are required for this change.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


tvararu (commented):
Just some comments about the env vars :+1:

.env.development
```diff
@@ -1,2 +1,4 @@
+DQT_API_URL=https://qualified-teachers-api-preprod.london.cloudapps.digital/
```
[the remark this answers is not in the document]
felixclack: This URL came from a conversation I had with James about which URL to use for development. He asked that we use this pre-prod one.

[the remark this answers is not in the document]
tvararu: Hmm ok!

app/controllers/trn_requests_controller.rb
```diff
@@ -9,18 +9,37 @@ def update
...
+      create_zendesk_ticket_and_redirect
+      return
+    end
+
+    begin
+      response = DqtApi.find_trn!(trn_request)
+      trn_request.update(trn: response['trn'])
+      redirect_to trn_found_path
+    rescue DqtApi::ApiError, Faraday::TimeoutError, DqtApi::TooManyResults
+      create_zendesk_ticket_and_redirect
```
tvararu, line 23: Do you this helper method is worth it when we're only duplicating it in two spots?

I think I would prefer to explicitly see the `redirect_to` calls inside this controller method. :thinking: 

When controller methods get long, I think the best pattern is to extract services/libs from what you can, and leave `redirect_to` and similar control flow calls inside the controller methods for clarity.

This is a non-blocking comment.

tvararu (commented):
Partial review, will continue!

app/services/feature_flag.rb
```diff
@@ -15,8 +15,11 @@ def feature
   PERMANENT_SETTINGS = [].freeze
 
   TEMPORARY_FEATURE_FLAGS = [
-    [:service_open, 'Allow users to access the service and submit TRN Requests', 'Theodor Vararu'],
+    [:dqt_api_always_timeout, 'Always time-out the DQT API', 'Felix Clack'],
+    [:dqt_api_too_many_results, 'Always return a too many results error from the DQT API', 'Felix Clack'],
```
tvararu, line 19: I think we don't need this feature flag in the end, we only really need one if we want to test the API going down. Should we remove it?

app/views/pages/trn_found.html.erb
```diff
@@ -0,0 +1,13 @@
+<% content_for :page_title, 'We’ve found your TRN' %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds-from-desktop">
+    <%= govuk_panel(
+      title_text: 'Success',
+      classes: %w[govuk-!-margin-bottom-6]
+    ) do %>
+      <h2 class="govuk-heading-m">We’ve found your TRN</h2>
```
tvararu, line 9:
```suggestion
      <h1 class="govuk-heading-m">We’ve found your TRN</h1>
```

spec/cassettes/DqtApi/_find_trn_/1_1_1.yml
```diff
@@ -0,0 +1,203 @@
...
+      method: get
+      uri: https://qualified-teachers-api-preprod.london.cloudapps.digital/v2/teachers/find?dateOfBirth&emailAddress=kevin@kevin.com&firstName=kevin&ittProviderName=provider&lastName&nationalInsuranceNumber=1000000
+      body:
+        encoding: US-ASCII
+        string: ''
+      headers:
+        User-Agent:
+          - Faraday v1.10.0
+        Authorization:
+          - Bearer kTWC7StTAeEKBpMz
```
tvararu, line 13: Shouldn't this be `Bearer testKey`? Were the tapes generated before we changed the `.env` files?

app/services/dqt_api.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    results = response.body['results']
+    raise TooManyResults if FeatureFlag.active?(:dqt_api_too_many_results) || results.size > 1
+
+    results.first
+  end
+
+  def self.trn_request_params(trn_request)
+    {
+      dateOfBirth: trn_request.date_of_birth,
```
tvararu, line 25: Does this serialise to `1990-01-01` without any `strftime` business? I didn't expect that, but that's cool if so

app/services/dqt_api.rb
```diff
@@ -0,0 +1,43 @@
...
+      emailAddress: trn_request.email,
+      firstName: trn_request.first_name,
+      ittProviderName: trn_request.itt_provider_name,
+      lastName: trn_request.last_name,
+      nationalInsuranceNumber: trn_request.ni_number,
```
tvararu, line 30: Should there be previous first/last names here as well?

spec/services/dqt_api_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    it { is_expected.to match(hash_including('trn' => '1275362')) }
+    it { is_expected.to match(hash_including('emailAddresses' => ['anonymous@anonymousdomain.org.net.co.uk'])) }
+    it { is_expected.to match(hash_including('firstName' => 'Kevin')) }
+    it { is_expected.to match(hash_including('lastName' => 'Evans')) }
+    it { is_expected.to match(hash_including('dateOfBirth' => '1990-01-01')) }
+    it { is_expected.to match(hash_including('nationalInsuranceNumber' => 'SD78771P9')) }
+    it { is_expected.to match(hash_including('uid' => 'f7891223-7661-e411-8047-005056822391')) }
+
+    context 'when the API returns a timeout error' do
```
tvararu, line 20: Hmm. I don't think we should test this functionality by using the `FeatureFlag`, as we're not exercising the path that the "real code" will take. (And if we were to use a coverage tool, it would highlight an untouched branch in the controller)

Could we instead force the API to fail in this spec using a mock?

app/controllers/trn_requests_controller.rb
felixclack: I did this because of a Rubocop warning. I can remove the helper method and modify the Rubocop rule instead.

app/services/dqt_api.rb
felixclack: Good point. They are missing.

spec/services/dqt_api_spec.rb
felixclack: I'll take a look at mocking a timeout

tvararu (approved):

PR #112, opened 2022-03-24
Switch to secure cookies in production

tvararu:
I couldn't verify this locally because it would require switching the protocol to HTTPS and a self-provisioned cert. But it should just work in Prod.


tvararu:
It worked!

image

PR #119, opened 2022-03-29
Prevent skipping steps

felixclack:
Currently, we don't enforce the order of the steps in the request and if
someone happens to know a valid URL, they can enter it and effectively
skip steps in the request flow.

The check answers screen doesn't expect missing values and throws an
error in those cases.

I've opted to go with a lightweight version of a state machine and
implement a `status` method on `TrnRequest`. This was the smallest step
I thought was needed to achieve our goal here.

This determines the furthest step that has been completed in the request
based on the presence of the attributes.

The system needs to allow people to go to a previously completed step as
well.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/concerns/enforce_question_order.rb
+++ b/app/controllers/concerns/enforce_question_order.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+module EnforceQuestionOrder
+  extend ActiveSupport::Concern
+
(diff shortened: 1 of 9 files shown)


tvararu (approved):
Tested locally, works really well! :shipit:

PR #126, opened 2022-04-01
Fix stylesheets

felixclack:
There's an issue where the styles defined in application.scss aren't
being picked up in the app.

The app setup isn't quite right and needs a few tweaks.

Firstly, we can't use both esbuild and importmap. Given the NPM packages
we need aren't guaranteed to work with ESM, we need to stick with
esbuild. Therefore, we can remove importmap.

Secondly, the styles for the GOVUK design system come from the NPM
package, govuk-frontend. These get compiled by esbuild and the output
file is application.css, which then overrides the application.css file
built by `rails-cssbundling`.

To allow both these styles and the app specific styles to co-exist, we
need to name the app stylesheet something other than `application.css`.

I opted for `main.css` to convey it is the main app stylesheet.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


bin/dev
```diff
@@ -1,3 +1,9 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
tvararu, line 1: Could this be `sh` instead because we might want to run this command inside `docker` which uses Alpine and doesn't come with bash preinstalled?

The `if` below might need to change :thinking:

tvararu (approved):
Could you also do a test that `docker build` produces a working container or at least successfully executes? Trying to make sure that we don't merge this and then it fails in the deploy pipeline :+1:

PR #128, opened 2022-04-01
Add Redis and unskip mailer test

felixclack:
Emailing a TRN requires we have Redis running. In order to test this on
CI, we need a Redis service.

The default GitHub setup is enough for our needs on CI.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


spec/system/trn_requests_spec.rb
```diff
@@ -308,18 +308,22 @@
...
       and_i_press_the_submit_button
       then_i_see_a_message_to_check_my_email
+      and_i_receive_an_email_with_the_trn_number
     end
   end
 
   private
 
+  def and_i_receive_an_email_with_the_trn_number
+    expect(ActionMailer::Base.deliveries.last.subject).to eq('Your TRN is 1275362')
```
tvararu, line 324: Cool! I saw Apply have some other helpers like `open_email` and `current_email`: https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/spec/system/candidate_interface/signup_and_signin/candidate_clicking_on_expired_magic_link_spec.rb#L48-L49

I think those come from `capybara-email`: https://github.com/DFE-Digital/apply-for-teacher-training/blob/f60a66a3f6e4e5f48d8372567569dec601c3a466/Gemfile#L145

Do you think it's worth adding them?

Non-blocking.

tvararu (approved):

spec/system/trn_requests_spec.rb
felixclack: Sure, happy to add them.

PR #141, opened 2022-04-13
Reset the user's session after they submit 

tvararu:
### Context

Once the user is presented with the TRN Found or Helpdesk request submitted page, we should reset their sessions so that they can't go back / edit their data and resubmit.

- Reset the user's session after they submit 
- Refactor update method in TrnRequests controller

### Guidance to review

Check that the tests make sense. Can review the `update` refactor separately as it's unrelated, I just felt like doing it.

### Checklist

https://trello.com/c/on7XJO4r/355-reset-a-users-session-after-they-submit-the-final-step

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/trn_requests_controller.rb
+++ b/app/controllers/trn_requests_controller.rb
@@ -7,26 +7,19 @@ class TrnRequestsController < ApplicationController
     session[:form_complete] = true
   end
 
-  def update # rubocop:disable Metrics/AbcSize
+  def update
     redirect_to root_url unless trn_request
     session[:form_complete] = false
 
     update_trn_request
-    unless FeatureFlag.active?(:use_dqt_api)
-      ZendeskService.create_ticket!(trn_request)
-      TeacherMailer.information_received(trn_request).deliver_now
-      redirect_to helpdesk_request_submitted_url
-      return
-    end
 
     begin
-      response = DqtApi.find_trn!(trn_request)
-      trn_request.update(trn: response['trn'])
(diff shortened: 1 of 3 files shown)


app/controllers/trn_requests_controller.rb
```diff
@@ -7,26 +7,19 @@ def show
...
-      TeacherMailer.found_trn(trn_request).deliver_now
+      find_trn_using_api
+
       redirect_to trn_found_path
     rescue DqtApi::ApiError, Faraday::ConnectionFailed, Faraday::TimeoutError, DqtApi::TooManyResults
-      ZendeskService.create_ticket!(trn_request)
-      TeacherMailer.information_received(trn_request).deliver_now
+      create_zendesk_ticket
+
       redirect_to helpdesk_request_submitted_url
```
tvararu, line 23: I think I previously was against writing something like this in a previous PR, but I think actually the thing that was irking me was bundling up the `redirect_to` inside the helper method as well. I think the control flow methods that are part of the controller should stay at the top level and be visible when just viewing the methods themselves, without drilling into helpers. Either way, curious if this makes sense.

felixclack: Makes sense to me.

felixclack: I think you're right about keeping the redirects in the action rather than the helper methods.

felixclack (approved):

PR #155, opened 2022-04-15
Create and use TrnDetailsComponent

tvararu:
### Context

Create a new component to share the logic for showing a user's answers across the:

- Check answers page
- No match page
- Support (anonymised)

### Changes proposed in this pull request

Bit of a grab bag / a bunch of stuff mixed in; individual commits are PR worthy on their own but I decided not to open 4 PRs as they're all related

- Add `factory_bot_rails` and `faker` with a factory for `trn_request`
- Create `TrnDetailsComponent`
- Add anonymised support (on by default in all `.production?` Rails modes)
- Display more info about TRNs and Zendesk tickets in Support

### Guidance to review

Review commit by commit.

#### Not yet submitted

image

#### TRN found

image

#### TRN not found, Zendesk not submitted

image

#### Zendesk submitted

image

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally

PR #156, opened 2022-04-18
Skip QTS question if we find a match eagerly

tvararu:
### Context

We want to let users skip over the ITT provider question if we're able to match their TRN using just the name, DOB, and NI number.

Query the API for the user's TRN earlier in the process and take the users to the email page if we find it.

### Guidance to review

Does not depend on https://github.com/DFE-Digital/find-a-lost-trn/pull/155, but that PR contains a change to the display logic for the check answers table to hide the ITT provider details if the user hasn't filled in the form with yes/no and the value is still `nil` in the database.

### Checklist

https://trello.com/c/Z4NVmIWx/341-skip-qts-question-if-we-find-a-match-with-just-name-dob-and-nino

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/dqt_api.rb
+++ b/app/services/dqt_api.rb
@@ -12,6 +12,8 @@ class DqtApi
   FIVE_SECONDS = 5
 
   def self.find_trn!(trn_request) # rubocop:disable Metrics/AbcSize
+    raise DqtApi::ApiError unless FeatureFlag.active?(:use_dqt_api)
+
     raise Faraday::TimeoutError, 'Time-out feature flag enabled' if FeatureFlag.active?(:dqt_api_always_timeout)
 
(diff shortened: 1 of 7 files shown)


app/services/dqt_api.rb
```diff
@@ -12,6 +12,8 @@ class NoResults < StandardError
   FIVE_SECONDS = 5
 
   def self.find_trn!(trn_request) # rubocop:disable Metrics/AbcSize
+    raise DqtApi::ApiError unless FeatureFlag.active?(:use_dqt_api)
```
felixclack, line 15: This is a better place to raise that error. 👏

felixclack (approved):
Nice job.

PR #160, opened 2022-04-19
Success panel improvements

tvararu:
https://trello.com/c/utA4lqIV/367-new-snags

### Before

image

### After

image

### DIFF
--- a/spec/components/trn_details_component_spec.rb
+++ b/spec/components/trn_details_component_spec.rb
@@ -40,7 +40,7 @@ RSpec.describe TrnDetailsComponent, type: :component do
   end
 
   it 'renders email' do
-    expect(component.text).to include('test@example.com')
+    expect(component.text).to include('test­@example­.com') # NB: This is test&shy;@example&shy;.com
   end
 
   it 'does not render change buttons' do
--- a/app/components/trn_details_component.rb
+++ b/app/components/trn_details_component.rb
@@ -48,7 +48,7 @@ class TrnDetailsComponent < ViewComponent::Base
     if @anonymise && !@trn_request.email.nil?
       "#{@trn_request.email.first}****@****.#{@trn_request.email.split('.').last}"
     else
-      @trn_request.email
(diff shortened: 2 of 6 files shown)


spec/components/trn_details_component_spec.rb
```diff
@@ -40,7 +40,7 @@
   end
 
   it 'renders email' do
-    expect(component.text).to include('test@example.com')
+    expect(component.text).to include('test­@example­.com')
```
tvararu, line 43: Bastard interviewer from hell: "Can you tell me what's different between these two lines?"

Good luck figuring it out without a hex editor.

The new code is `test&shy;@example&shy;.com`, but instead of `&shy;` it's the actual unicode glyph for the soft hyphen character. It appears as just a space in my editor and my terminal (diff between an actual `%20` space and the new glyph):

image

I'm pretty sure at some point this will bite us if we leave it as is.

I tried doing this instead, but it didn't work:

```
expect(component.text).to include(ApplicationHelper.shy_email('test@example.com'))
```

What could I try instead @felixclack?

felixclack: I'm uncertain about the value of this particular change. 
Am I correct in assuming that we want to allow people to enter email addresses with spaces in them, and then preserve those spaces using the soft hyphen character to pass validations?

felixclack (approved):
I think overall the changes look solid. Sanitising the email address is great. The fixes for the layout are also a good improvement.

Not sure about the soft-hyphens but there's a separate comment for that and it's not a blocker for me.

spec/components/trn_details_component_spec.rb
tvararu: No, there's a different issue here.

The email is still "test@example.com", but it's being output as "test&shy;@example&shy;.com".

As such, I had to update the assertion in the test, but I think this is unintuitive because the characters I had to add are rendered as invisible on some editors and on GitHub.

Maybe I'll just add a comment about this.

PR #162, opened 2022-04-20
Handle support TRN details when email is nil

tvararu:
Fixes https://sentry.io/organizations/dfe-teacher-services/issues/3209850541/?project=6275068&referrer=slack

### DIFF
--- a/app/components/trn_details_component.rb
+++ b/app/components/trn_details_component.rb
@@ -45,7 +45,9 @@ class TrnDetailsComponent < ViewComponent::Base
   end
 
   def email
-    if @anonymise && !@trn_request.email.nil?
+    return 'Not provided' if @trn_request.email.nil?
+
+    if @anonymise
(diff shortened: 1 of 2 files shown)


app/components/trn_details_component.rb
```diff
@@ -45,7 +45,9 @@ def itt_provider_value
   end
 
   def email
-    if @anonymise && !@trn_request.email.nil?
+    return 'Not provided' if @trn_request.email.nil?
+
+    if @anonymise
```
felixclack, line 50: Minor thing that isn't blocking... I like the pattern of returning early that you did here. It removes branching logic and is easier to extend in the future. I would also do the same for the other return values for the method. I find it makes reading the code a lot easier.

```ruby
  def email
    return 'Not provided' if @trn_request.email.nil?
    return "#{@trn_request.email.first}****@****.#{@trn_request.email.split('.').last}" if @anonymise
    
    helpers.shy_email(@trn_request.email)
  end
```

felixclack (approved):

PR #163, opened 2022-04-20
Ensure name_changed is nil if prev names are nil

tvararu:
### Before

image

### After

image


fofr (approved):

felixclack (approved):
Very neat change. Much better experience for the user.

PR #164, opened 2022-04-20
Add a question about being awarded a QTS

felixclack:
As part of the lessons learned from user feedback, we want to separate
out the question of whether someone has been awarded QTS from how they
were awarded it.

This allows us to be more inclusive of people who were awarded a QTS outside a
school or university.

I opted to introduce a new attribute on `TrnRequest` to store the answer
to the question of if someone has a QTS.

This allows us to use the `itt_provider_enrolled` attribute to now mean
whether someone was able to supply the name of the training provider or
not.

| New question | New answers |
| - | - |
| Screen Shot 2022-04-20 at 1 36 14 pm | Screen Shot 2022-04-20 at 1 41 42 pm |

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -2,7 +2,9 @@
 
 module ApplicationHelper
   def back_link_url(back = url_for(:back))
-    return check_answers_path if session[:form_complete] && request.path != check_answers_path
+    if session[:form_complete] && [check_answers_path, itt_provider_path].exclude?(request.path)
+      return check_answers_path
+    end
 
     back
   end
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -41,3 +41,7 @@ RSpec/ExampleLength:
 RSpec/MultipleExpectations:
   Exclude:
     - spec/**/*
+
+RSpec/MultipleMemoizedHelpers:
+  Exclude:
+    - spec/components/trn_details_component_spec.rb
--- a/app/components/trn_details_component.html.erb
+++ b/app/components/trn_details_component.html.erb
@@ -25,6 +25,12 @@
     actions: [{ href: have_ni_number_path, visually_hidden_text: 'national insurance number' }]
   }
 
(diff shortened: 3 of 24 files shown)


tvararu (commented):
Found a few of bugs when checking locally:

image

:bug: Bug 1: Click `Change awarded QTS` link and select "No", it takes me to the email page instead of the review page:

image

:bug: Bug 2: Following from above, when I get to the check answers page, it shows the old University SCITT value or the "No" value for the second form:

image
image

Possible solution for bug 1: Double check where the form takes the user after submitting, make sure it's check answers when appropriate.

Possible solution for bug 2: Set the values to `nil` for the `IttProviderForm` fields when `AwardedQtsForm` sets `awarded_qts` to `nil`.

---

:bug: Bug 3: Click `Change awarded QTS` link, select `Yes`, it takes you to the ITT provider page. The back button takes you to `Check answers` instead of the previous page.

app/helpers/application_helper.rb
```diff
@@ -2,7 +2,9 @@
 
 module ApplicationHelper
   def back_link_url(back = url_for(:back))
-    return check_answers_path if session[:form_complete] && request.path != check_answers_path
+    if session[:form_complete] && [check_answers_path, itt_provider_path].exclude?(request.path)
+      return check_answers_path
+    end
```
tvararu, line 7: I think this isn't working correctly; found a :bug: bug:

Review page -> Click "Change teacher training provider" -> `itt_provider_path` -> Click back -> :exclamation: `awarded_qts_path`

Instead if should go to `check_answers_path`.

Having a think how to fix this...

felixclack: 🤔

felixclack: Added a fix for this.

tvararu (approved):

PR #168, opened 2022-04-25
Log validation errors

felixclack:
We want visibility into how people are using the service and in
particular, which error messages they see.

I took inspiration from [Apply](https://github.com/DFE-Digital/apply-for-teacher-training/pull/1922) when designing this.

Storing validation errors as a record in the database makes it easy for
us to display in the support interface. The UI for this will come in a
future change.

The scope of where a validation error can occur is more limited than the
Apply version, so I was able to keep it simpler.

An error will always be associated with a `TrnRequest` and the form
object that was validated has a one-to-one mapping with a particular
path, so we know where the error was seen.

I considered using a controller concern but instead went for including
it at the form object level. The idea being that logging the errors is
the responsibility of the form object that triggered the errors rather
than having another object have knowledge of both the form object and
how to record the errors.

Apply made the choice to log in the controller to avoid logging
validation errors more than necessary, ie. they assumed that form
objects would be used in more places than the controller.

In our case, we have an assumption that a form object is only used
by the controllers. We have no other way to use them. The pattern we
have followed is to use the form objects to encapsulate all side-effects
related to creating/mutating objects in the system.

This means that controller actions are responsible for co-ordinating
this process but not the implementation of it.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/.env.development
+++ b/.env.development
@@ -1,5 +1,6 @@
 DQT_API_URL=https://preprod-teacher-qualifications-api.education.gov.uk/
 DQT_API_KEY=testkey
+GOVUK_NOTIFY_API_KEY=test
 HOSTING_ENVIRONMENT_NAME=local
 SUPPORT_PASSWORD=test
 SUPPORT_USERNAME=test
--- a/.env.development
+++ b/.env.development
@@ -1,6 +1,5 @@
 DQT_API_URL=https://preprod-teacher-qualifications-api.education.gov.uk/
 DQT_API_KEY=testkey
-GOVUK_NOTIFY_API_KEY=test
 HOSTING_ENVIRONMENT_NAME=local
 SUPPORT_PASSWORD=test
 SUPPORT_USERNAME=test
--- a/app/models/concerns/log_errors.rb
+++ b/app/models/concerns/log_errors.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+module LogErrors
+  extend ActiveSupport::Concern
+
+  included do
+    def log_errors
+      return unless FeatureFlag.active?(:log_validation_errors)
+
+      ValidationError.create(form_object: self.class.name, messages: errors.messages, trn_request: trn_request)
+    end
+  end
+end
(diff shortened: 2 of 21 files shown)


.env.development
```diff
@@ -1,5 +1,6 @@
 DQT_API_URL=https://preprod-teacher-qualifications-api.education.gov.uk/
 DQT_API_KEY=testkey
+GOVUK_NOTIFY_API_KEY=test
```
tvararu, line 3: I wrote this in the README about getting your own Notify key:

https://github.com/DFE-Digital/find-a-lost-trn#notify

Then you'd actually be able to open the Notify dashboard and see the emails that were sent out. If you use a dummy key, the mailer will silently fail right?

So I guess the question here is whether for development we want to blow up and ask that the user sets up a local test key, or silently fail.

What's your preference?

felixclack: I have a test key as well. 

Personally, I don't think we should have the development .env in version control. I end up stashing changes to it for every new branch I work on.

tvararu: You can have `.env.development.local`, does that work?

felixclack: Yeah, ok. I'll use that.

tvararu: Cool, then yeah I think we can keep `.local` ones out of version control and the `.env.development` can stay in as that fits the suggested use here https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use

tvararu, line 3:
```suggestion
```

Let's remove this as per above discussion

app/models/concerns/log_errors.rb
```diff
@@ -0,0 +1,12 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    def log_errors
+      return unless FeatureFlag.active?(:log_validation_errors)
+
+      ValidationError.create(form_object: self.class.name, messages: errors.messages, trn_request: trn_request)
+    end
+  end
+end
```
tvararu, line 12: Nice and simple this!

app/services/feature_flag.rb
```diff
@@ -16,6 +16,7 @@ def feature
 
   TEMPORARY_FEATURE_FLAGS = [
     [:dqt_api_always_timeout, 'Always time-out the DQT API', 'Felix Clack'],
+    [:log_validation_errors, 'Log validation errors', 'Felix Clack'],
```
tvararu, line 19: Good thinking to add a feature flag for this :+1:

tvararu (approved):

PR #170, opened 2022-04-27
Add a UI for viewing validation errors

felixclack:
This is the first iteration of a support interface for viewing the
validation errors.

Initially, the intention is to display a list representing the total
errors for each of the form objects.

A show page then provides the details of each validation error for that
form object.

Assumptions:
* we can add pagination in a follow-up PR
* search would be useful but can also be added in a follow-up PR

Screen Shot 2022-04-27 at 2 08 43 pm
Screen Shot 2022-04-27 at 2 08 53 pm

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/assets/stylesheets/main.sass.scss
+++ b/app/assets/stylesheets/main.sass.scss
@@ -50,3 +50,7 @@ $govuk-new-link-styles: true;
     text-align: right;
   }
 }
+
+.app-p0 {
+  padding: 0;
+}
--- a/app/views/layouts/base.html.erb
+++ b/app/views/layouts/base.html.erb
@@ -46,6 +46,11 @@
           text: "Sidekiq",
           href: support_interface_sidekiq_web_path
         ) %>
+        <% header.navigation_item(
+          active: current_page?(support_interface_validation_errors_path),
+          href: support_interface_validation_errors_path,
+          text: 'Validation errors'
+        ) %>
       <% end %>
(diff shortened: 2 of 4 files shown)


spec/system/performance_spec.rb
```diff
@@ -2,7 +2,7 @@
 require 'rails_helper'
 
 RSpec.describe 'Performance', type: :system do
-  before { Timecop.freeze(Date.new(2022, 0o4, 21)) }
+  before { Timecop.freeze(Date.new(2022, 04, 21)) }
```
tvararu, line 5: Lol how did this work

tvararu (approved):
Nice! Can you add a link to the support header to this section?

spec/system/performance_spec.rb
felixclack: I assume that Date is very forgiving. 😄

fofr:
A few design comments:

Can the date and time format be:
`27 April 2022 at 1:04pm`

Can we add a link to the validation pages in the header

On the child pages, after clicking "NameForm", that page should have a breadcrumb to navigate back up to the top menu: `Validation errors > NameForm`

The title of each page does not need the caption.

Can we remove the top margin/padding on the summary list so that `attribute` lines up with the request ID.

PR #173, opened 2022-05-03
Post weekly performance stats in a Slack channel

felixclack:
We want to provide an easy way to track the performance metrics from
Find a Lost TRN.

It was decided that a message in Slack was the most appropriate way of
doing this.

We followed the lead of the teacher vacancies project to implement the scheduling
piece of this work with [sidekiq-cron](https://github.com/ondrejbartas/sidekiq-cron).

Right now, the message we post is very simple and only contains the
total number of TrnRequests created in the last week.

To complete setup of this, a Slack web hook URL is required,
ie. SLACK_WEBHOOK_URL.

This configuration variables allow us to run this in multiple environments and verify
it works.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


adr/00007-scheduled-jobs.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Context
+
+This service will need to run jobs on a schedule. We need a way to run domain logic code outside the context of web requests, reliably and in a proactive/scheduled way.
+
+### Findings
+
+We have considered:
+
+- Find-style: run jobs via the infrastructure (they use sidekiq/sidekiq_cron)
```
tvararu, line 17:
```suggestion
- run jobs via the infrastructure (they use sidekiq/sidekiq_cron)
```

Find-style is a relic of `Find postgraduate teacher training`.

adr/00007-scheduled-jobs.md
```diff
@@ -0,0 +1,73 @@
...
+- clockwork gem on its own
+
+Considering the ability to enforce time-dependent business rules is core functionality
+for the service, we have decided to follow best practices and implement scheduling
+and background processing properly, using a similar approach to Find, but using
```
tvararu, line 22:
```suggestion
and background processing properly, but using
```

tvararu (commented):
Maybe instead of duplicating the content of the Apply scheduling ADR, we can just link to it and say "We will follow Apply's reasoning and use the `clockwork` gem: \<link\>"

app/services/slack_client.rb
```diff
@@ -0,0 +1,32 @@
...
+    return if response.success?
+
+    raise Error, "Status code: #{response.status} - #{response.body}"
+  end
+
+  private
+
+  def client
+    Faraday.new(url: 'https://slack.com/api') do |faraday|
+      faraday.authorization :Bearer, ENV['SLACK_TOKEN']
```
tvararu, line 25: Can you add the `SLACK_TOKEN` environment variable to the terraform files and relay it to @merrilmathew who can add it to the environments?

config/clock.rb
```diff
@@ -0,0 +1,13 @@
...
+require './config/environment'
+require 'clockwork'
+
+class Clock
+  include Clockwork
+
+  configure { |config| config[:tz] = 'Europe/London' }
+  error_handler { |error| Sentry.capture_exception(error) if defined?(Sentry) }
+
+  every(1.week, 'performance.alert', at: 'Monday 09:00') { PerformanceAlertJob.perform_later }
```
tvararu, line 12: Should we only run this in the production hosting environment?

tvararu (commented):
Before we merge this, can you test it by running the worker with a production Slack key from your local environment and @-ing me the resulting message on Slack? :+1:

saliceti:
Could we please consider NOT using clock. It doesn't play well with terraform so much that we have to delete and recreate it at each deployment: https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/.github/actions/deploy/action.yml#L96

Teacher services have aligned on https://github.com/ondrejbartas/sidekiq-cron instead. It's easy to use (example: https://github.com/DFE-Digital/teaching-vacancies/blob/main/config/schedule.yml) and uses standard sidekiq workers

tvararu:
@saliceti thanks for the suggestion!

app/services/slack_client.rb
```diff
@@ -0,0 +1,32 @@
...
+  def self.create_message(message)
+    new.create_message(message)
+  end
+
+  def create_message(message)
+    response =
+      client.post(
+        'chat.postMessage',
+        { channel: "##{ENV['SLACK_CHANNEL']}", mrkdwn: true, text: message },
+      )
```
tvararu, line 15: This uses the Slack API to post a message to a certain channel, which requires us to provision a Slack user with full read/write access.

There's a simpler way using the [Slack webhook API](https://api.slack.com/messaging/webhooks). It has less privileges; it only fires a message into a channel and can't read anything. Creating the Webhook API URL key should also be easier.

Here's how Apply does it: https://github.com/DFE-Digital/apply-for-teacher-training/blob/e85ff42219275d6728efabb225464340b195178d/app/workers/slack_notification_worker.rb#L7

tvararu: Also, we can just hardcode the `SLACK_CHANNEL` and not set it in an environment variable.

tvararu (changes requested):

Gemfile
```diff
@@ -64,6 +65,7 @@ end
 group :test do
   gem 'capybara', '~> 3.36'
   gem 'capybara-email'
+  gem 'clockwork-test', '~> 0.4.0'
```
tvararu, line 68: Think this can be removed if we're not using Clockwork anymore?

app/services/slack_client.rb
```diff
@@ -0,0 +1,27 @@
...
+class SlackClient
+  class Error < StandardError
+  end
+
+  def self.create_message(message)
+    new.create_message(message)
+  end
+
+  def create_message(message)
+    response = client.post('', { text: message })
```
tvararu, line 11:
```suggestion
    response = client.post('/', { text: message })
```

I think this is just a little bit clearer, wdyt?

tvararu (approved):
Smashing!!! :rocket:

PR #204, opened 2022-06-07
Update accessibility statement

tvararu:
Add note about the JAWS bug we are currently investigating.

Because we've added a "Non-accessible content" section, [according to the guidance in the sample statement](https://www.gov.uk/government/publications/sample-accessibility-statement/sample-accessibility-statement-for-a-fictional-public-sector-website), we must also add a "Disproportionate burden" section and a "Content that’s not within the scope of the accessibility regulations" section, even if we have nothing to report for them.

https://trello.com/c/Qn1eluqa/424-arrange-internal-accessibility-audit-address-findings

### DIFF
--- a/app/views/static/accessibility.md
+++ b/app/views/static/accessibility.md
@@ -71,6 +71,40 @@ accordance with the Public Sector Bodies (Websites and Mobile Applications)
 This service is fully compliant with [the Web Content Accessibility Guidelines
 version 2.2 AA standard](https://www.w3.org/TR/WCAG22/).
 
+## Non-accessible content
+
+The content listed below is non-accessible for the following reasons.
+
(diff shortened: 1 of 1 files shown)


tvararu:
Drafting while we investigate fixing the underlying issue a bit more.

app/views/static/accessibility.md
```diff
@@ -85,9 +119,9 @@ any changes we make to address any issues raised.
 ## Preparation of this accessibility statement
 
 This statement was prepared on Tuesday 5 April 2022. It was last reviewed on
-Tuesday 5 April 2022.
+Tuesday 7 June 2022.
 
 This service was last tested on Thursday 24 March 2022. The test was carried
```
fofr, line 124: Does the tested date need to change?

PR #205, opened 2022-06-13
Enable retrieving data from Zendesk

felixclack:
When a TRN Request cannot be found via the API, a Zendesk ticket is
created. Often, this manual intervention results in a matching TRN being
found.

We want to be able to retrieve this TRN from the Zendesk ticket and then
use it to lookup the details of the teacher via the API and then create
a diff of the details provided with the details returned by the API.

The aim is for us to be able to learn how we can adapt the service to
resolve these scenarios without a manual intervention.

The first pass at this implements a mechanism to retrieve the TRN from
Zendesk and also retrieve the response from the DQT API. However, it is
required to be triggered manually by pressing a button in the support
interface.

The automation of this process can be implemented in a future change.

Screen Shot 2022-06-14 at 1 07 22 pm

### Anonymised

Screen Shot 2022-06-14 at 1 06 34 pm

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


tvararu (commented):
Excellent work :+1: 

Only thing I think this needs to be approved is doing the `trn_request.update(trn:)` using the TRN we get from the DQT API instead of the RegExp. It's a subtle difference but I think it's worth it.

Rest are up for debate or can be done in follow-up PRs.

PR #208, opened 2022-06-20
Switch to using prettier.js for everything

felixclack:
We can run the Ruby prettier formatting using the JS library. This
unifies our project and means we can format both Ruby and JS using the
same command.

Config for Ruby needs to happen via SyntaxTree plugins. Currently only
single quotes and trailing comma formatting are supported.

Everything else comes from Rubocop.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history

### DIFF
--- a/app/mailers/teacher_mailer.rb
+++ b/app/mailers/teacher_mailer.rb
@@ -17,4 +17,12 @@ class TeacherMailer < ApplicationMailer
 
     notify_email(mailer_options)
   end
+
+  def delayed_information_received(trn_request)
+    @trn_request = trn_request
+
+    mailer_options = { to: @trn_request.email, subject: 'We’ve received the information you submitted' }
+
+    notify_email(mailer_options)
+  end
 end
--- a/app/views/teacher_mailer/delayed_information_received.erb
+++ b/app/views/teacher_mailer/delayed_information_received.erb
@@ -1,0 +1,20 @@
+Dear <%= @trn_request.name %>,
+
+You recently requested help with finding your teacher reference number (TRN). We’re sorry you did not get an acknowledgement at the time. This was due to a technical problem, which we’ve now resolved.
+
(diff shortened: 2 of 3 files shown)


.prettierrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "printWidth": 120,
+  "singleQuote": true,
+  "bracketSpacing": true,
+  "trailingComma": "es5",
+  "rubyPlugins": "plugin/single_quotes,plugin/trailing_comma"
+}
```
tvararu, line 7: Interesting!

.github/workflows/lint.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           skip-node: true
 
       - name: Run Prettier
-        run: bin/bundle exec rbprettier --check --ignore-unknown --loglevel warn '**/*'
+        run: npx prettier --check --ignore-unknown --loglevel warn '**/*'
```
tvararu, line 24: Let's use the yarn version of this instead of relying on `npx`, which is part of `npm`:

```suggestion
        run: yarn run prettier --check --ignore-unknown --loglevel warn '**/*'
```

.github/workflows/lint.yml
```diff
@@ -37,7 +37,9 @@ jobs:
           skip-node: true
```
tvararu, line 37: Removing this will prevent us from explicitly running `yarn` below:

```suggestion
```

.github/workflows/lint.yml
```diff
@@ -37,7 +37,9 @@ jobs:
...
 
       - name: Run Rubocop
-        run: bin/bundle exec rubocop
+        run: |
+          yarn
```
tvararu, line 41: See above about removing `skip-node: true`:

```suggestion
```

tvararu (approved):

PR #209, opened 2022-06-20
Enable iterative processing of Zendesk comments

felixclack:
When searching for a TRN in the comments of a Zendesk ticket there is an
assumption that the last comment contains the matching text.

In practice, this isn't always true. A ticket can have multiple comments
and the last one isn't guaranteed to contain the TRN.

This change will iteratively process the comments, starting from the
second one (the first is always a copy of the original request). The
search will stop as soon as the regex matches.

I opted for this as the simplest improvement we could make to the
current process. It still makes an assumption that the regex will
eventually match. This assumption may not always be true in practice.

We are looking into ways of making this more reliable in the future.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,11 +17,9 @@ jobs:
 
       - name: Prepare application environment
         uses: ./.github/actions/prepare-app-env
-        with:
-          skip-node: true
 
       - name: Run Prettier
-        run: bin/bundle exec rbprettier --check --ignore-unknown --loglevel warn '**/*'
+        run: yarn prettier --check --ignore-unknown --loglevel warn '**/*'
 
   rubocop:
     name: Rubocop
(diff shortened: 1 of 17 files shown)


app/services/fetch_trn_from_zendesk.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    update_from_zendesk
+  end
+
+  private
+
+  def trn
+    @trn ||=
+      zendesk_ticket.comments[1..]
+        .find { |comment| comment.body.scan(/\*\*(\d{7})\*\*/).flatten.first }
```
tvararu, line 23: I think the regular expression could be a variable, and reused below

tvararu (approved):

PR #210, opened 2022-06-20
Configure Sidekiq to run in production

felixclack:
This is a direct port from
[Apply](https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/110).

It sets up a single worker to run with the same configuration as the web app.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


tvararu (approved):
Let's merge this on its own, after we merge + deploy everything else. When we deploy, we should look at the logs to make sure everything is set up correctly.

PR #219, opened 2022-06-23
Fix posting to Slack

felixclack:
When sending a message to a Slack webhook, a 302 response is returned.
This happens because of the trailing slash in post command.

There is no reason to pass a slash character in the post command, so we
safely remove it.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


tvararu (approved):
Good spot!

PR #220, opened 2022-06-23
Customise the performance alert URL per host

felixclack:
The URL that is included in the performance alert is hard-coded to
production. This makes it hard to distinguish alerts from other
environments.

We occasionally need to test this alert in other environments, so it is
beneficial to allow us to customise the URL.

I opted to use an environment variable to store the current hosting
domain, that can then be passed to the Rails URL helper to generate the
appropriate value.

This seems simpler and easier to extend across environments than the
alternative of keeping a list of all expected domains in code somewhere.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,6 +1,7 @@
 DQT_API_URL=https://preprod-teacher-qualifications-api.education.gov.uk/
 DQT_API_KEY=testkey
 GOVUK_NOTIFY_API_KEY=test
+HOSTING_DOMAIN=http://test
 HOSTING_ENVIRONMENT_NAME=test
 SLACK_WEBHOOK_URL=test
 SUPPORT_USERNAME=test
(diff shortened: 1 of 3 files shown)


tvararu (approved):
Domains added to env vars:

```bash
find-a-lost-trn $ make dev print-keyvault-secret | grep HOSTING_DOMAIN
HOSTING_DOMAIN: https://dev-find-a-lost-trn.education.gov.uk
find-a-lost-trn $ make test print-keyvault-secret | grep HOSTING_DOMAIN
HOSTING_DOMAIN: https://test-find-a-lost-trn.education.gov.uk
find-a-lost-trn $ make preprod print-keyvault-secret | grep HOSTING_DOMAIN
HOSTING_DOMAIN: https://preprod-find-a-lost-trn.education.gov.uk
find-a-lost-trn $ make production print-keyvault-secret | grep HOSTING_DOMAIN
HOSTING_DOMAIN: https://find-a-lost-trn.education.gov.uk
```

PR #237, opened 2022-06-28
Performance code refactor

benilovj:
### Context

[We want to iterate the performance dashboard to produce better insight.](https://trello.com/c/uuopi3dY/234-iterate-the-performance-dashboard) but the existing code doesn't really make extension straightforward.

Tidy the existing performance code, following the patterns set in [Apply for QTS](https://github.com/DFE-Digital/apply-for-qualified-teacher-status):

- introduce the `PerformanceController` for rendering the `/performance` endpoint
- extract a `PerformanceStats` class to encapsulate the calculation of the stats
- add in lower-level, more granular spec to reduce the risk of regression


tvararu (approved):
Thanks a lot for this, Jake!

PR #239, opened 2022-06-30
Handle a Zendesk API failure

felixclack:
We had an incident where the connection to the Zendesk API failed and
we were unable to create tickets for requests that had no match.

This resulted in failed requests and manual work to resolve the issue.

We discussed solutions for this and decided that using our job queue
would allow us to gracefully handle this and retry the requests to
Zendesk at a later time to ensure they still get created.

If the API fails, we inform the person requesting their TRN that there
will be a delay while we address the issue. Then we queue a job to keep
trying to create the ticket.

The default Sidekiq retry schedule ensures that the job will be retried
at exponentially increasing intervals. This should give the API time to
recover or, in the case where manual intervention is required, we have
a mechanism to retry the job manually.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

Screenshot 2022-07-01 at 11 02 08 am


app/controllers/trn_requests_controller.rb
```diff
@@ -30,9 +30,15 @@ def update
...
            DqtApi::TooManyResults,
            TrnHasAlert
-      create_zendesk_ticket
-
-      redirect_to helpdesk_request_submitted_path
+      begin
+        create_zendesk_ticket
+        redirect_to helpdesk_request_submitted_path
+      rescue ZendeskService::ConnectionError => exception
+        CreateZendeskTicketJob.set(wait: 1.day).perform_later(trn_request.id)
```
tvararu, line 37: 1 day seems like a long wait before retrying again, in the case of something like a small flake in the API. Could it be maybe 5 minutes, with exponential backoff? (You mentioned it should do this by default in the PR description)

config/routes.rb
```diff
@@ -76,6 +76,7 @@
 
   get "/ask-questions", to: "pages#ask_questions"
   get "/helpdesk-request-submitted", to: "pages#helpdesk_request_submitted"
+  get "/delayed-zendesk-ticket", to: "pages#delayed_zendesk_ticket"
```
tvararu, line 79: I don't think we should make `zendesk-ticket` a part of our URLs, because it's an implementation detail as far as the user is concerned. Instead, we could try:

```suggestion
  get "/helpdesk-request-delayed", to: "pages#helpdesk_request_delayed"
```

tvararu (commented):
Looks good, just some suggestions + will approve after a content review :+1:

app/jobs/create_zendesk_ticket_job.rb
```diff
@@ -0,0 +1,12 @@
+class CreateZendeskTicketJob < ApplicationJob
+  discard_on ActiveRecord::RecordNotFound
+
+  def perform(trn_request_id)
+    trn_request = TrnRequest.find(trn_request_id)
+    return if trn_request.zendesk_ticket_id
+
+    ZendeskService.create_ticket!(trn_request)
+    TeacherMailer.information_received(trn_request).deliver_now
```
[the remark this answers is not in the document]
tvararu: Nope, let's keep the email as is!

tvararu (approved):

PR #246, opened 2022-07-04
Control Slack alerts via a feature flag

felixclack:
We want to enable the Slack alerts on a per environment basis.

Rather than hardcoding these environment settings in code, we can use
a feature flag to control them. This gives us the ability to test them
in environments other than production without having to make a code
change.

### Checklist

- [ ] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/feature_flag.rb
+++ b/app/services/feature_flag.rb
@@ -13,7 +13,11 @@ class FeatureFlag
   end
 
   PERMANENT_SETTINGS = [
-    [:slack_alerts, "Allow sending alerts to Slack", "Felix Clack"]
+    [
+      :slack_alerts,
+      "Enable Slack alerts and notifications for this environment",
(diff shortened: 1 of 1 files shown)


app/services/feature_flag.rb
```diff
@@ -13,7 +13,7 @@ def feature
   end
 
   PERMANENT_SETTINGS = [
-    [:zendesk_health_check, "Check Zendesk health", "Felix Clack"]
+    [:slack_alerts, "Allow sending alerts to Slack", "Felix Clack"]
```
tvararu, line 16:
```suggestion
    [:slack_alerts, "Enable Slack alerts and notifications for this environment", "Felix Clack"]
```

tvararu (approved):

PR #261, opened 2022-07-11
Explicitly depend on view_component

benilovj:
### Context

We have a couple of view components in the codebase that rely on the `view_component` gem. Currently, that gem is pulled into the app indirectly view the `govuk_components` gem.

Declare an explicit runtime dependency on `view_component` in `Gemfile`. It's highly unlikely that we'll ever be taking out `govuk_components` as a dependency but it's useful for new devs to see the dependency explicitly as an indicator that we use view components.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


tvararu (approved):
Thank you! :+1:

PR #278, opened 2022-07-14
Avoid loading Rails for the smoke test

felixclack:
There's an issue when the smoke test is called in a hosted environment,
it tries to load Rails but fails.

We don't want to load the application in the smoke test as we are
trying to test the hosted version.

This change switches the smoke test to not load the application.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history

### DIFF
--- a/spec/system/smoke_spec.rb
+++ b/spec/system/smoke_spec.rb
@@ -1,9 +1,11 @@
 # frozen_string_literal: true
-require "faker"
-require "shoulda-matchers"
-require "rails_helper"
+require "spec_helper"
+require "capybara/rspec"
+require "capybara/cuprite"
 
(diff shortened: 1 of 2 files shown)


spec/system/smoke_spec.rb
```diff
@@ -1,9 +1,27 @@
...
+    process_timeout: 30,
+    window_size: [1200, 800],
+    debug: true,
+    logger: $stdout,
+    timeout: 60,
+  )
+end
+Capybara.default_driver = :cuprite
+Capybara.javascript_driver = :cuprite
+Capybara.run_server = false
```
tvararu, line 20: Could this be `smoke_test_helper` or such?

tvararu (approved):

PR #280, opened 2022-07-14
Performance page: journeys through the service

benilovj:
### Context

We want to see what journeys our users are taking through the service. We have new designs for this (see the [bottom panel in the prototype](https://find-a-lost-trn-prototype.herokuapp.com/performance)).

Trello: https://trello.com/c/uuopi3dY/415-iterate-the-performance-dashboard
Previous PRs: #247 #262 #273

Add a new table to the performance page, which breaks down complete journeys through the app

image

### Guidance to review

This PR deviates from linked design in a couple of places:

- the table caption has been updated (after a discussion on Slack)
- the table doesn't split out the cases where a user went back to edit their details, as I'm not sure how to identify those in the data.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


tvararu:
> the table doesn't split out the cases where a user went back to edit their details, as I'm not sure how to identify those in the data.

Hmm, indeed, we don't store that data yet

app/components/performance_table_helpers.rb
```diff
@@ -0,0 +1,24 @@
...
+          100 * attributes[numerator_key].fdiv(attributes[:total]),
+          precision: 0
+        )
+      "(#{percentage})"
+    end
+  end
+
+  def number_with_percentage_cell(counts, key, label: nil)
+    [
+      number_with_delimiter(counts[key]),
```
tvararu, line 17: I discovered there's actually an ActiveSupport built-in with almost the same name that does the same thing https://api.rubyonrails.org/classes/ActiveSupport/NumberHelper.html#method-i-number_to_delimited

We should refactor away our custom one, not in this PR necessarily

tvararu (approved):

tvararu (approved):

PR #282, opened 2022-07-14
Performance page: fix double counting in TRN request classification

benilovj:
## What's the issue?

This commit fixes a double-counting bug in the "TRN requests by day" and the "TRN requests by month" sections on the performance page.

image

In the example above, for July 2022, the sum of the "TRN found", "No match" and "Did not finish" columns is 3,122 requests, but the total column shows 2,931 requests.

## Root cause

For a TRN request that was initially unsuccessful, a Zendesk ticket ID will be populated. Later, when the helpdesk finds the TRN for this user, the TRN will also be populated on the record via the background Zendesk scanning job.

These requests were being double-counted in both the "TRN found" and "No match" categories.

### DIFF
--- a/app/lib/performance_stats.rb
+++ b/app/lib/performance_stats.rb
@@ -40,21 +40,27 @@ class PerformanceStats
 
   private
 
+  def arel_columns_for_request_counts
+    [
+      Arel.sql(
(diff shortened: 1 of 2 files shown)


tvararu (approved):

tvararu:
image
image

benilovj:
Lol @tvararu

benilovj:
image

PR #287, opened 2022-07-20
General Refactor and Documentation Updates

ransom4real:
### Context

We do not catch missing env vars earlier in the app loading hence we get failures during initialisation without useful error messages with help on how to fix. Also the documentation for testing and running the app is missing a few bits and bobs

### Changes proposed in this pull request

- Update app initialisation to catch missing env vars earlier than later with a clearer failure message and reference to appropriate README sections
- Add precompile script to test section of readme
- Updated build css script to quiet warnings
- Added README section for handling basic auth where feature flag for service open is set to false

### Guidance to review

Verify documentation is accurate and useful
Review error message for missing env var to ensure it is clear

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history
- [ ] Tested by running locally

PR #292, opened 2022-07-21
Throw a better message if Notify key is missing

tvararu:
Previously was done in https://github.com/DFE-Digital/find-a-lost-trn/pull/287, but that implementation included a bug that wasn't detected in the review: the value for `notify_settings` must be a hash with an `api_key` field.

Tested locally; Notify sends emails successfully with the new initializer code:

image


tvararu:
Going to test this on the review app as well.

benilovj (approved):

tvararu:
Works via Review app too:

image

PR #293, opened 2022-07-22
Updated ITT Provider form to use auto complete

rmdasilvaD2l:
### Context

The ITT provider field field needs to be changed from a text field to an auto complete.

This pull request updates the ITT providers text field to use an auto complete using the DfE Autocomplete (https://github.com/DFE-Digital/dfe-autocomplete)

### Guidance to review

The changes are used behind a feature flag called `use_dqt_api_itt_providers`.

This is not a final pull request and is only used to show current progress before making any further design and implementation decisions.


rmdasilvaD2l:
Re-raised using a new branch at https://github.com/DFE-Digital/find-a-lost-trn/pull/303

PR #294, opened 2022-07-22
Match on Email

ransom4real:
### Context

We want to be able to match TRN records on the Date of Birth step instead of the NI step

- There is a feature flag to turn on matching on the DOB step instead of the NI step. This is because until qualified-teachers-api ships an update on their side, if we shipped this, it would probably break all "eager matching." It also allows us to control when this is enabled and see how it impacts our performance stats
- Matching is done at the date of birth step instead of the NI step
- Don't include email when querying DQT when feature flag is off

### Guidance to review

- Check New Tests
- Test on Review

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


tvararu (commented):
Code looks good so far; let's add a system spec that exercises the path with the feature flag on. Then we can merge this :rocket: 

We can go back and update the `dqt_api_spec` once we remove the feature flag as a follow-up.

ransom4real:
> Code looks good so far; let's add a system spec that exercises the path with the feature flag on. Then we can merge this 🚀
> 
> We can go back and update the `dqt_api_spec` once we remove the feature flag as a follow-up.

This is now done and in the branch. Should be good to merge in once i merge changes from main into this branch

tvararu (approved):
:clap:

PR #310, opened 2022-07-28
Use Sidekiq to deliver emails

felixclack:
We want to send emails using a background job to make them retryable in
the case of failure.

I considered writing a custom job to handle this but opted instead to
keep things simple and use the default setup.

Calling `deliver_later` queues a `ActionMailer::MailDeliveryJob` to
perform sending the email.

I opted to keep using the `default` queue for these jobs as we don't
have a large volume of jobs where separate queues would help with
prioritisation.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,21 +11,33 @@ help: ## Show this help
 dev:
 	$(eval DEPLOY_ENV=dev)
 	$(eval AZURE_SUBSCRIPTION=s165-teachingqualificationsservice-development)
+	$(eval RESOURCE_NAME_PREFIX=s165d01)
+	$(eval ENV_SHORT=dv)
+	$(eval ENV_TAG=dev)
 
 .PHONY: test
 test:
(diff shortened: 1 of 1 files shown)


tvararu:
:+1: Going to test this on the review app before merging/shipping

tvararu:
image

Confirmed, still works!

tvararu (approved):

spec/system/performance_spec.rb
```diff
@@ -41,7 +41,7 @@ def when_i_visit_the_performance_page
   end
 
   def then_i_see_the_live_stats
-    expect(page).to have_content("36\nrequests today and the previous 7 days")
+    expect(page).to have_content("36 requests today and the previous 7 days")
```
tvararu, line 44: Seems like this is failing the tests?

PR #319, opened 2022-07-29
Make staff users invitable

tvararu:
Similar to https://github.com/DFE-Digital/apply-for-qualified-teacher-status/pull/143

Depends on https://github.com/DFE-Digital/find-a-lost-trn/pull/318

This makes staff users invitable.

[Trello card](https://trello.com/c/op2Sq8d9/591-integrate-a-real-user-authentication-system-for-the-support-section)


ransom4real (approved):
Looks great. Approving. However noticed there were no model specs, is this intentional?

tvararu:
> Looks great. Approving. However noticed there were no model specs, is this intentional?

Yep; we're not really using much custom code in our models, just devise calls. I don't think testing them is particularly valuable, they're pretty battle tested already.

PR #322, opened 2022-08-01
Unlock Teacher Self Service Account

ransom4real:
### Context

In the manual support process, a support agent would unlock a user's Teacher Self Service Portal Account when they request their TRN.

We want to do the same for users when they receive a TRN via Find a lost TRN.

[Trello card](https://trello.com/c/YVXdflHt/579-unlock-a-teachers-self-service-portal-account-when-we-find-their-trn)

- There is a feature flag to turn this feature on or off
- Implement API call to the https://qualified-teachers-api-tech-docs.london.cloudapps.digital/#v2-unlock-teacher-teacherid end point
- Handle error states and degrade gracefully to prevent calls to this end point from breaking find_trn! call in the situation it finds a TRN. Log errors to Sentry
- Update current `with a matching TRN request` spec to ensure `unlock_teacher` method is not called when feature flag is not active

### Guidance to review

- Ensure happy (Successful request) and sad (Api Error, Timeouts e.t.c) paths are covered
- Test on review

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


app/services/dqt_api.rb
```diff
@@ -27,7 +27,21 @@ def self.find_trn!(trn_request)
...
+    if FeatureFlag.active?(:unlock_teachers_self_service_portal_account)
+      begin
+        DqtApi.unlock_teacher(teacher_account.fetch("uid"))
+      rescue KeyError,
+             ApiError,
+             Faraday::ConnectionFailed,
+             Faraday::TimeoutError,
+             TooManyResults => e
+        Sentry.capture_exception(e)
+      end
```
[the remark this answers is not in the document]
ransom4real: @felixclack makes absolute sense. Will effect this change now.

app/services/dqt_api.rb
```diff
@@ -44,6 +48,25 @@ def self.get_itt_providers
...
+          new.client.put(
+            "/v2/unlock-teacher/#{teacher_account.fetch("uid")}",
+            {}
+          )
+        raise ApiError, response.reason_phrase unless response.success?
+      rescue KeyError,
+             ApiError,
+             Faraday::ConnectionFailed,
+             Faraday::TimeoutError,
+             TooManyResults => e
```
tvararu, line 64: I don't think we can get `TooManyResults` from this endpoint:

```suggestion
             Faraday::TimeoutError => e
```

tvararu (commented):
Nice one. Can you add a unit test for the `.unlock_teacher!` method?

app/services/dqt_api.rb
ransom4real: Great spot @tvararu . Will update and crack on with unit tests for the `.unlock_teacher!` method

spec/services/dqt_api_spec.rb
```diff
@@ -84,4 +84,87 @@
...
+        {
+          trn: "2921020",
+          emailAddresses: ["anonymous@anonymousdomain.org.net.co.uk"],
+          firstName: "Kevin",
+          lastName: "Evans",
+          dateOfBirth: "1990-01-01",
+          nationalInsuranceNumber: "AA123456A",
+          uid: "f6891223-7661-e431-8047-005056822391"
+        }
+      end
```
tvararu, line 126: If only the `uid` is different, can that be a `let(:uid)` and leave us with only one `let(:teacher_account)` hoisted at the top level that references the `uid`?

Also, can we make the non-matching UID look more obviously wrong? Like `a-non-matching-uid`.

spec/services/dqt_api_spec.rb
```diff
@@ -84,4 +84,87 @@
...
+        {
+          trn: "2921020",
+          emailAddresses: ["anonymous@anonymousdomain.org.net.co.uk"],
+          firstName: "Kevin",
+          lastName: "Evans",
+          dateOfBirth: "1990-01-01",
+          nationalInsuranceNumber: "AA123456A",
+          uid: "f689122-7661-e431-8047-00506822391"
+        }
+      end
```
tvararu, line 141: See above comment

spec/services/dqt_api_spec.rb
```diff
@@ -84,4 +84,87 @@
...
+        {
+          trn: "2921020",
+          emailAddresses: ["anonymous@anonymousdomain.org.net.co.uk"],
+          firstName: "Kevin",
+          lastName: "Evans",
+          dateOfBirth: "1990-01-01",
+          nationalInsuranceNumber: "AA123456A"
+        }
+      end
+      it { is_expected.to be_nil }
```
tvararu, line 156: See above, here we can make `uid: nil`

tvararu (commented):
Looks good! I think the `let`s in the tests can be cleaned up a bit and the commits as well. Then it's good to go!

spec/services/dqt_api_spec.rb
ransom4real: @tvararu we want to verify a KeyError is raised if the expected key is not present in the returned data. Setting this to nil will only pass an empty value but what we want is for the KeyError to be raised. Correct me if i am wrong

tvararu (approved):

ransom4real:
Closing PR as i messed up the commit logs somehow after a rebase. Cutting new branch and raising new PR

PR #325, opened 2022-08-01
Add an OKComputer check for Notify

felixclack:
We want to check the health of the Notify integration as part of the
overall system health check.

There doesn't appear to be an endpoint specifically for testing the
connection, so I opted to follow the guidance for creating a smoke test.

Notify has special handling for specigic smoke testing email addresses.
This allows us to test the API integration without sending a real email.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


tvararu:
Nice
image

tvararu (approved):

PR #335, opened 2022-08-02
Unlock teacher self service portal account

ransom4real:
### Context

In the manual support process, a support agent would unlock a user's Teacher Self Service Portal Account when they request their TRN.

We want to do the same for users when they receive a TRN via Find a lost TRN.

[Trello card](https://trello.com/c/YVXdflHt/579-unlock-a-teachers-self-service-portal-account-when-we-find-their-trn)

- There is a feature flag to turn this feature on or off
- Implement API call to the https://qualified-teachers-api-tech-docs.london.cloudapps.digital/#v2-unlock-teacher-teacherid end point
- Handle error states and degrade gracefully in the `unlock_teacher!` method to prevent breaking `find_trn!` call in the situation it finds a TRN. Log errors to Sentry

### Guidance to review

- Ensure happy (Successful request) and sad (Api Error, Timeouts e.t.c) paths are covered
- Test on review

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/spec/services/dqt_api_spec.rb
+++ b/spec/services/dqt_api_spec.rb
@@ -85,4 +85,60 @@ RSpec.describe DqtApi do
       end
     end
   end
+
+  describe ".unlock_teacher!" do
+    before do
+      FeatureFlag.activate(:unlock_teachers_self_service_portal_account)
+    end
+    after do
+      FeatureFlag.deactivate(:unlock_teachers_self_service_portal_account)
+    end
+
+    subject(:unlock_teacher!) do
+      described_class.unlock_teacher!(teacher_account)
+    end
+    let(:uid) { "f7891223-7661-e411-8047-005056822391" }
+    let(:teacher_account_base) do
+      {
+        trn: "2921020",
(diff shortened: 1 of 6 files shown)


spec/services/dqt_api_spec.rb
```diff
@@ -85,4 +85,60 @@
...
+
+    context "when teacher ID is not found", vcr: true do
+      let(:uid) { "f6891223-7661-e431-8047-005056822391" }
+      it { is_expected.to be_nil }
+    end
+
+    context "when teacher ID is not valid", vcr: true do
+      let(:uid) { "a-non-matching-uid" }
+      it { is_expected.to be_nil }
+    end
```
tvararu, line 126: Aren't these testing the same thing fundamentally?

tvararu (approved):

spec/services/dqt_api_spec.rb
ransom4real: Not found returns a 404 but the format of the ID is valid. The next returns a 400 indicating the format of the ID is invalid. So i think we are technically testing different things.

PR #345, opened 2022-08-04
Scrub all environment variable values from Sentry logs

ransom4real:
### Context

In our last incident, Sentry logged an error in Slack that contained the content of our Notify key, thus leaking it: https://sentry.io/organizations/dfe-teacher-services/issues/3444625248/?referrer=slack

Add proc to ActiveSupport ParameterFilter instance to substitute all occurrences of environment variables in sentry logs with [FILTERED]

### Guidance to review

- Enable Sentry Logging on local by adding the appropriate SENTRY_DSN
- Add `SENSITIVE_KEY=donotleakthis` to your `.env.development.local` file so we can verify it is scrubbed
- In the `app/controllers/pages_controller.rb` `start` method, `raise ENV["SENSITIVE_KEY"]` exception
- Check Sentry scrubs `donotleakthis` from the Stack trace across the board

### Checklist

- [x] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


config/initializers/sentry.rb
```diff
@@ -1,9 +1,16 @@
...
   filter =
     ActiveSupport::ParameterFilter.new(
-      Rails.application.config.filter_parameters
+      Rails.application.config.filter_parameters +
+        [
+          ->(k, v) do
+            return unless k == "value" && k == "title"
+            v.gsub!(env_var_filters, "[FILTERED]") if env_var_filters.match?(v)
+          end
+        ]
```
tvararu, line 13: I think this could look a bit nicer if it was a variable and we hoisted it a bit higher.

Actually think this could be `env_var_filter` and the union could be called like `env_var_values`

Just stylistic nit, non-blocking

tvararu (approved):
Nice work!

config/initializers/sentry.rb
ransom4real: The union returns a Regexp object and it is not necessarily the values anymore hence why i have named it env_var_filters as it contains an array of Regexp's

ransom4real:
@tvararu @felixclack Rubocop and Prettier are at loggerheads over `do`..`end` vs `{..}` and they seem to contradict each other hence causing my PR to be blocked.

config/initializers/sentry.rb
```diff
@@ -1,9 +1,24 @@
...
 Sentry.init do |config|
   config.environment = HostingEnvironment.environment_name
+  filter_parameters = Rails.application.config.filter_parameters.map(&:to_s)
+  env_var_filters =
+    Regexp.union(
+      ENV
+        .select { |k, _v|
+          filter_parameters.any? { |parameter| k.downcase.include?(parameter) }
+        }
+        .values
```
tvararu, line 11: The linters are arguing because we're chaining a select that takes a proc with a `.values` call. You can work around this by using separate variables instead of chaining.

I don't think we need this code however; it wasn't here for my last review.

I understand that it's pre-emptively filtering out all environment variables whose keys don't contain any of the `filter_parameters` we've defined.

This however would miss out on a bunch of secret-y env vars, like:

- `SENTRY_DSN`
- `VCAP_SERVICES`
- `DATABASE_URL`
- `SLACK_WEBHOOK_URL`

Why is it necessary to select from the env vars in this way?

config/initializers/sentry.rb
tvararu: Right; could we still hoist it? If we do, it makes sense to name it something like `env_filter` in which case we should name the Regexp union something more relating to values. Naming things in this context is a bit difficult because the Regexp is a filter, the `ActiveSupport::ParameterFilter` is a filter, and things get confusing.

What do you think of this; at the least, I much prefer putting the proc in its own variable:

```ruby
  env_values = Regexp.union(ENV.values)
  env_filter = ->(k, v) do
    return unless %i[value title].include?(k)
    v.gsub!(env_values, "[FILTERED]") if env_values.match?(v)
  end
  filter =
    ActiveSupport::ParameterFilter.new(
      Rails.application.config.filter_parameters + [env_filter]
    )
```

tvararu (changes requested):

config/initializers/sentry.rb
ransom4real: On my local host I noticed it was picking up a lot of irrelevant ENV vars causing the Regexp to act erratically when used with the .gsub! method. I suspect this would also produce erratic behaviour on other devs localhost. This was to allow us narrow down our own env vars rather than everything contained in that localhost.

ransom4real: I could add _services and _url to the filter params list and it should pick the others. I already added _dsn earlier

config/initializers/sentry.rb
ransom4real: Makes absolute sense. Will effect this change

tvararu (approved):
Nice :+1:

PR #348, opened 2022-08-05
Match against DQT after NINO question, even when email match is enabled.

benilovj:
### Context

This is a suspected bug, relating to matching on email ([Trello](https://trello.com/c/h59LbSMJ/560-match-on-email)).

When email matching is enabled, the current behaviour skips a lookup to DQT API after the NINO question.

This forces a user to have to unnecessarily answer the ITT question, even in the case when they'd get a match with name, DOB and NINO.

Change the logic to always match against DQT API after the NINO question, irrespective of whether email matching is enabled or not.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/ni_number_controller.rb
+++ b/app/controllers/ni_number_controller.rb
@@ -14,16 +14,14 @@ class NiNumberController < ApplicationController
       session[:ni_number_not_known] = true
       next_question
     elsif @ni_number.update(ni_number_params)
-      unless FeatureFlag.active?(:match_on_email)
-        begin
-          find_trn_using_api unless @trn_request.trn
-        rescue DqtApi::ApiError,
-               Faraday::ConnectionFailed,
-               Faraday::TimeoutError,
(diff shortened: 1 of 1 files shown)


benilovj:
Tested this change on the review app. In the scenario:

* the email doesn't match
* name matches
* DOB matches
* NINO matches

the ITT provider is (correctly) no longer asked.

ransom4real (commented):
Are we removing the flag usage all together? If so then is there a need for the flag to be used in the temporary flag constant list?

tvararu (approved):

tvararu:
> Are we removing the flag usage all together? If so then is there a need for the flag to be used in the temporary flag constant list?

@ransom4real we still need it in the `date_of_birth_controller`; this change just makes it so that if it's enabled, we attempt to match both on the `dob` step as well as the `ni` step.

PR #362, opened 2022-08-08
DQT API Request Client Retry

ransom4real:
### Context

Calls to the [DQT unlock teacher account](https://trello.com/c/YVXdflHt) is slow and timeout yet we do not currently attempt to retry

Add Unlock Account Action Job to Retry on Faraday::TimeoutError until successful Asynchronously

### Guidance to review

Verify Unlock Account Action Job Test Coverage

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


ransom4real:
> Did you consider moving the unlock action into a background job? This would handle the retries for you on failure and also without increasing the response time for the find_trn! method.

Makes absolute sense. I will effect this change right away

app/services/dqt_api.rb
```diff
@@ -55,8 +55,10 @@ def self.unlock_teacher!(teacher_account)
             {}
           )
         raise ApiError, response.reason_phrase unless response.success?
-      rescue ApiError, Faraday::ConnectionFailed, Faraday::TimeoutError => e
+      rescue ApiError, Faraday::ConnectionFailed => e
         Sentry.capture_exception(e)
+      rescue Faraday::TimeoutError
+        UnlockTeacherAccountJob.perform_later(teacher_account)
```
tvararu, line 61: Maybe we should add a short delay to this retry? Maybe after 5 minutes?

app/jobs/unlock_teacher_account_job.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+class UnlockTeacherAccountJob < ApplicationJob
```
tvararu, line 2: Now that we have this, we can replace the call to `DqtApi.unlock_teacher!(teacher_account)` with `UnlockTeacherAccountJob.perform_later(teacher_account)` in `DqtApi`; this will make the `find_trn!` action quicker, as it will make the unlock teacher action run asynchronously :+1:

tvararu (approved):
Looks good!

app/services/dqt_api.rb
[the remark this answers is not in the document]
ransom4real: How does that work if you are not calling the Job to perform a deferred task? I assumed you will have to reference the Job at some point for it to retry

app/services/dqt_api.rb
```diff
@@ -27,7 +27,7 @@ def self.find_trn!(trn_request)
 
     teacher_account = results.first
 
-    DqtApi.unlock_teacher!(teacher_account)
+    UnlockTeacherAccountJob.set(wait: 5.minutes).perform_later(teacher_account)
```
tvararu, line 30: This shouldn't run after 5 minutes on the initial run, you can run it immediately after, could you open a follow-up PR?

PR #364, opened 2022-08-09
Add a data migration to encrypt existing records

felixclack:
Any record that was stored in the database prior to the addition of
encryption still has sensitive values in plain text.

This data migration runs on all unencrypted records and performs the
encryption to change them.

The `encrypt` method updates all fields on a record marked for
encryption. It doesn't modify any timestamps, so the `updated_at` field
can still be relied on to represent the last user update.

I considered only encrypting records that hadn't been encrypted yet.
Either by checking the encryption status of all the required fields or
by using a timestamp of when we deployed the change.

Given the volume of records we have is low and the chances of missing
some data with whatever approach we take is non-zero, it seems safest to
simply encrypt every record in the DB. Fields that are already encrypted
will be re-encrypted.

The cost of this is that it extends the running time of the migration.

I believe this is an acceptable cost for the certainty that all records
have been encrypted correctly.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


db/data/20220809085018_encrypt_data.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class EncryptData < ActiveRecord::Migration[7.0]
+  def up
+    TrnRequest.find_each(&:encrypt)
```
tvararu, line 5: Nice

tvararu (approved):

gunndabad:
[Identity support UI](https://trello.com/c/g87c1EYy/532-identity-support-ui)

PR #366, opened 2022-08-09
Remove the feature flag use_dqt_api

felixclack:
We added a feature flag to control using the DQT API to retrieve a TRN.

This was introduced at a time when we were building out the
functionality.

The flag has been enabled for enough time to give us confidence that we
wouldn't want to turn this feature off in the future.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/controllers/itt_providers_controller.rb
```diff
@@ -15,9 +15,7 @@ def update
     if @itt_provider_form.save
       next_question
     else
-      if FeatureFlag.active?(:use_dqt_api_itt_providers)
-        set_itt_provider_options
-      end
+      set_itt_provider_options
```
tvararu, line 18: Ah, this is a different feature flag; we haven't switched this on yet. Please leave it in :+1:

tvararu (changes requested):

tvararu (approved):

PR #380, opened 2022-08-11
Add host to hosting_environment

tvararu:
This was missed in https://github.com/DFE-Digital/find-a-lost-trn/pull/378

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -43,15 +43,6 @@ module FindALostTrn
 
     config.exceptions_app = routes
 
-    config.action_mailer.notify_settings = {
-      api_key:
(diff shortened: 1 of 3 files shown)


config/application.rb
```diff
@@ -43,15 +43,6 @@ class Application < Rails::Application
 
     config.exceptions_app = routes
 
-    config.action_mailer.notify_settings = {
```
ransom4real, line 46: Is this no longer needed? @tvararu

tvararu: Nope; we're only using Notify in production now, so no need to specify a key when developing locally

ransom4real (approved):

PR #383, opened 2022-08-11
Verify parameters from Identity

rmdasilvaD2l:
### Context

Identity will post parameters to the Identity endpoint. The parameters will include a signature that can be used to check if the parameters are correct and have not been modified upon reaching the Find A Lost TRN service.

Checks that the parameters passed from Identity are valid by creating a HMAC signature and checking that it matches the supplied signature from Identity.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


rmdasilvaD2l:
[/identity page: Parameters and HMAC verification](https://trello.com/c/7TYRCrHZ/619-identity-page-parameters-and-hmac-verification)

app/controllers/identity_controller.rb
```diff
@@ -1,15 +1,76 @@
...
+  def valid_identity_params?
+    return false if expected_signature.blank? || new_signature.blank?
+
+    expected_signature.upcase == new_signature.upcase
+  end
+
+  # Retrieve the list of params required for the paramater check. The keys
+  # must be in alphabetical order.
+  def sorted_identity_params
+    params_without_signature = allowed_create_params.except("sig")
```
gunndabad, line 38: See the note at https://github.com/DFE-Digital/get-an-identity/blob/main/docs/find-lost-trn-integration.md#computing-the-signature
Ideally this needs to look at all params (not just the ones that are currently known about) for future compatibility.

rmdasilvaD2l: Thanks @gunndabad I did discuss this with @tvararu and if the params are unlikely to be updated very often we would like to use permitted params to have more control of what can data passed in. Is this something that is likely to change much in the future?

gunndabad: It will almost certainly change at some point. We can keep as-is for now and review in future.

PR #396, opened 2022-08-17
Add console1984

felixclack:
We want a way of tracking console sessions. console1984 provides this
with sensible defaults.

It is configured to request a name when starting a session, along with a
purpose for the session.

All the commands entered are stored in the new tables and we will
introduce a companion library, audits1984, in a future change to be able
to review them.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2022-08-17 at 10 05 49 am
Screenshot 2022-08-17 at 10 06 04 am
Screenshot 2022-08-17 at 10 06 25 am
Screenshot 2022-08-17 at 10 06 40 am
Screenshot 2022-08-17 at 10 06 51 am

### DIFF
--- a/adr/00009-delete-zendesk-tickets.md
+++ b/adr/00009-delete-zendesk-tickets.md
@@ -1,0 +1,45 @@
+# 9. Delete Zendesk tickets
+
+Date: 2022-08-16
+
+## Status
+
+Accepted
+
+## Context
+
+We need to delete Zendesk tickets that are 6 months or older due to our data
+retention policy. Some of these Zendesk tickets are created by Find, and others
+are created through other means.
+
+Find is in a good position to handle deleting tickets, because it already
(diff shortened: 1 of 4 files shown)


db/migrate/20220817082352_create_console1984_tables.console1984.rb
```diff
@@ -0,0 +1,37 @@
...
+      t.index :created_at
+      t.index %i[user_id created_at]
+    end
+
+    create_table :console1984_users do |t|
+      t.string :username, null: false
+      t.timestamps
+
+      t.index [:username]
+    end
```
tvararu, line 18: Can we make this somehow reference staff users instead? :thinking:

tvararu (approved):
Can merge this in and iterate the user auth later

PR #406, opened 2022-08-22
Send the TRN back to Identity and redirect if the initial request was from Identity

rmdasilvaD2l:
### Context

After being redirected to Find A TRN from Identity and a TRN has been found, send an API request back to Identity with the TRN and then redirect to the supplied redirect uri.

Creates a new API call to Identity to send the TRN for an Identity journey. Changes the controller logic to send the TRN back to Identity and redirect instead of continuing the Find A TRN process if the initial request came from Identity. If no matching TRN was found an API request will still be made with no TRN.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/no_match_controller.rb
+++ b/app/controllers/no_match_controller.rb
@@ -13,9 +13,26 @@ class NoMatchController < ApplicationController
     if @no_match_form.valid? && @no_match_form.try_again?
       redirect_to check_answers_path
     elsif @no_match_form.valid?
-      create_zendesk_ticket
+      if trn_request.from_get_an_identity?
+        begin
+          # Send a request to Get An Identity API with no TRN found
+          IdentityApi.submit_trn!(trn_request, session[:identity_journey_id])
 
-      redirect_to helpdesk_request_submitted_path
+          # Send the user back to Get an Identity
+          redirect_to session[:identity_redirect_uri], allow_other_host: true
+        rescue IdentityApi::ApiError,
+               Faraday::ConnectionFailed,
+               Faraday::TimeoutError,
+               ActionController::ActionControllerError => e
+          Sentry.capture_exception(e)
+          # Do not redirect back to Get An Identity
+          render "errors/internal_server_error", status: :internal_server_error
+        end
+      else
+        create_zendesk_ticket
+
+        redirect_to helpdesk_request_submitted_path
+      end
     else
       render :new
     end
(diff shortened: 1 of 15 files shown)


rmdasilvaD2l:
[POST /check-answers: Handover from Find to Get an identity](https://trello.com/c/V2KF46Rp/621-post-check-answers-handover-from-find-to-get-an-identity)

spec/system/identity_spec.rb
```diff
@@ -25,6 +25,43 @@
     then_i_see_the_check_answers_page
   end
 
+  context "when sending the trn to the Identity API" do
+    it "redirects back to Get An Identity", vcr: true do
+      and_i_access_the_identity_endpoint_with_parameters
```
tvararu, line 30: Can we add an `alias_method` so this reads better?

```suggestion
      when_i_access_the_identity_endpoint_with_parameters
```

app/controllers/no_match_controller.rb
```diff
@@ -13,9 +13,25 @@ def create
...
 
-      redirect_to helpdesk_request_submitted_path
+          # Send the user back to Get an Identity
+          redirect_to session[:identity_redirect_uri], allow_other_host: true
+        rescue IdentityApi::ApiError,
+               Faraday::ConnectionFailed,
+               Faraday::TimeoutError,
+               ActionController::ActionControllerError
+          # Do not redirect back to Get An Identity
+          render "errors/unprocessable_entity", status: :unprocessable_entity
```
tvararu, line 28: I don't know if 422 is correct here; unprocessable entity means that the request was understood, is well formed, but the server couldn't handle it *and it is the client's fault in some way.* If the downstream API fails, it isn't the client's fault. It also generally means that the client should not repeat the request without any changes, as it'll just be another 422, which isn't the case here.

Probably more of a 500 scenario + a sentry alert.

app/controllers/trn_requests_controller.rb
```diff
@@ -22,9 +22,24 @@ def update
...
 
-      redirect_to trn_found_path
+          # Send the user back to Get an Identity
+          redirect_to session[:identity_redirect_uri], allow_other_host: true
+        rescue IdentityApi::ApiError,
+               Faraday::ConnectionFailed,
+               Faraday::TimeoutError,
+               ActionController::ActionControllerError
+          # Do not redirect back to Get An Identity
+          render "errors/unprocessable_entity", status: :unprocessable_entity
```
tvararu, line 37: See above

tvararu (approved):

PR #408, opened 2022-08-22
Implement the ask for TRN page + check answers section

ransom4real:
### Context

[Trello Card](https://trello.com/c/kPHnncq0)

> Get an Identity is a single sign-on service that will hand over users to Find in order to enrich their profiles with additional information, such as their Names or TRN.

During the Get an identity journey, we want to ask users what their TRN is in case they already know it.

We should implement the page for this from the prototype:

https://find-a-lost-trn-prototype.herokuapp.com/account/trn-known

- The ask for TRN page exists
- The page comes after the NI number but before the ITT provider
- The route is `/ask-trn`
- It implements the design from the prototype
- We have validation in place
  - `presence`, and `length: { maximum: 7 }`, digits only
  - Add basic validation messages too
  - Strip away non-digit characters
- There is a check answers section to update a user's supplied TRN
- The user submitted TRN is saved differently from the TRN returned by the API
- Ask for TRN page is not shown if we match earlier in the process

### Guidance to review

All of the above criteria's have been met and tests cover both happy and sad path scenarios

### Ask for TRN Page

image

#### No options selected

image

#### Empty TRN input 

image

#### TRN input less than 7 digits

image

#### TRN input more than 7 digits

image

#### I know my TRN

image

#### I do not know my TRN

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/components/trn_details_component.rb
```diff
@@ -91,4 +91,10 @@ def previous_name
       @trn_request.previous_name
     end
   end
+
+  def trn_from_user_value
+    return @trn_request.trn_from_user if session[:user_knows_trn] == "true"
+
+    "I don’t know my TRN"
```
tvararu, line 98: Could you provide a screenshot in the PR description with what this looks like so that our designers can give it a look?

app/controllers/ask_trn_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def new
+    @ask_trn_form = AskTrnForm.new(trn_request:)
+    @ask_trn_form.do_you_know_your_trn = session[:user_knows_trn]
+  end
+
+  def create
+    @ask_trn_form = AskTrnForm.new(trn_request:)
+    if @ask_trn_form.update(trn_params)
+      session[:user_knows_trn] = trn_params.require(:do_you_know_your_trn)
```
tvararu, line 15: Could we remove the need for this session variable? They're undesirable as they act like global variables, and make testing harder.

In this case, could we set the `trn_from_user` database field to `""` when the users specify they do not know their TRN, and then use that value at later points?

app/components/trn_details_component.html.erb
```diff
@@ -48,6 +54,7 @@
   rows.push(previous_name_row) if @trn_request.previous_name?
   rows.push(date_of_birth_row)
   rows.push(ni_row)
+  rows.push(ask_trn_row) if FeatureFlag.active?(:ask_for_trn)
```
tvararu, line 57: This row shouldn't be visible if the user isn't in the identity flow.

Could we check for that as well? The attribute is `.from_get_an_identity` on a `TrnRequest`.

app/controllers/concerns/enforce_question_order.rb
```diff
@@ -95,6 +96,14 @@ def ask_for_ni_number?
     trn_request.has_ni_number && trn_request.ni_number.nil?
   end
 
+  def ask_for_trn?
+    return false unless FeatureFlag.active?(:ask_for_trn)
```
tvararu, line 100: We also shouldn't ask the user for their TRN if they are not in the Identity flow; see above comment.

app/forms/ask_trn_form.rb
```diff
@@ -0,0 +1,38 @@
...
+    if invalid?
+      log_errors
+      return false
+    end
+
+    trn_request.save!
+  end
+
+  def trn_from_user=(value)
+    trn_request.trn_from_user = value&.delete("^0-9")
```
tvararu, line 32: Nice

spec/system/trn_requests_spec.rb
```diff
@@ -598,6 +599,26 @@ def given_the_use_dqt_api_itt_providers_feature_is_enabled
...
+    FeatureFlag.deactivate(:match_on_email)
+  end
+
+  def given_the_ask_for_trn_feature_is_enabled
+    FeatureFlag.activate(:ask_for_trn)
+  end
+
+  def given_the_ask_for_trn_feature_is_disabled
+    FeatureFlag.deactivate(:ask_for_trn)
+  end
```
tvararu, line 620: Are these being used?

tvararu (commented):
Added some comments, also needs a system spec :+1:

spec/system/trn_requests_spec.rb
ransom4real: It was to be used for the system specs which i am working on

app/services/feature_flag.rb
```diff
@@ -58,6 +58,11 @@ def feature
       :unlock_teachers_self_service_portal_account,
       "Unlock a teacher's self service portal account if their TRN is found",
       "Ransom Voke Anighoro"
+    ],
+    [
+      :ask_for_trn,
+      "Ask users for their TRN if they already know it",
+      "Ransom Voke Anighoro"
```
tvararu, line 65: This feature actually doesn't need a feature flag, because it's part of the identity work. We can use that feature flag instead.

ransom4real: Do you mean `from_get_an_identity` ?

tvararu: Yes

app/components/trn_details_component.rb
ransom4real: This has now been provided in the PR description

spec/system/trn_requests_spec.rb
ransom4real: No they are not. Taking them out

app/services/feature_flag.rb
ransom4real: Removed

app/controllers/ask_trn_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  include EnforceQuestionOrder
+
+  layout "two_thirds"
+
+  def new
+    @ask_trn_form = AskTrnForm.new(trn_request:)
+    @ask_trn_form.do_you_know_your_trn = "true" if @ask_trn_form
+      .trn_request
+      .trn_from_user
+      .present?
```
tvararu, line 12: Could this logic sit in the initializer for `AskTrnForm`?

tvararu (commented):
Excellent work!

Last thing I think would be useful, could you add a new system spec that tests this form at a basic level?

app/controllers/ask_trn_controller.rb
ransom4real: @tvararu not sure as it is not an attribute of the `trn_request` model. What do you suggest?

ransom4real:
> Excellent work!
> 
> Last thing I think would be useful, could you add a new system spec that tests this form at a basic level?

@tvararu This is ready to review. Added tests to verify it is included as part of the Identity journey and that it is not part of non Identity journeys. Including happy and sad paths

tvararu (approved):
:raised_hands:

PR #426, opened 2022-08-25
Implement custom layout with header displaying the `client_title` from Get an identity

ransom4real:
[Trello Card](https://trello.com/c/svzIPqc8)

### Context

> Get an Identity is a single sign-on service that will hand over users to Find in order to enrich their profiles with additional information, such as their Names or TRN.

When users reach Find a lost TRN from Get an identity, Get an identity will give us a `client_title` param.

We should change the header title in Find to this.

- Add `client_title` from Identity journey to session
- Create `custom_title` helper
- Replace title header ERB with `custom_title` method

### Guidance to review

- Users on the Get an identity journey see a customised title
- Make sure this doesn't introduce XSS or security vulnerabilities
  - There should be some tests testing evil input

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/identity_controller.rb
+++ b/app/controllers/identity_controller.rb
@@ -28,6 +28,7 @@ class IdentityController < ApplicationController
     session[:trn_request_id] = @trn_request.id
     session[:identity_journey_id] = allowed_create_params["journey_id"]
     session[:identity_redirect_uri] = allowed_create_params["redirect_uri"]
+    session[:client_title] = allowed_create_params["client_title"]
 
     redirect_to next_question_path
   end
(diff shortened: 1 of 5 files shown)


tvararu:
Can you please customise the `govuk_header` too?

Can we add some tests too, especially for evil input?

ransom4real:
> Is there a system test you could write here as well? Seems like we're only testing the helper in this change, not the actual process of assigning the title from the session.

You are right @felixclack . I assumed it would be duplication of effort since the helper is used in exactly the same way? If you reckon it is still needed it shouldn't be difficult to add one happy and one evil path system test

ransom4real:
@tvararu @felixclack all suggested changes have been made.

PR #442, opened 2022-09-01
Customise govuk_header with title from Identity

tvararu:
The `govuk_header` service name should be bespoke during the Identity journey, so that users feel like they are going through a single connected service.

Also ensure the `page_title` comes up prefixed in the actual `<title>`.

I couldn't test this end to end locally and coming up with a solution for that was taking too much time, so I opted to skip on those improvements for now.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -31,13 +31,11 @@ module ApplicationHelper
   end
 
   def custom_title(page_title)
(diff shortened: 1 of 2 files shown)


ransom4real (approved):

ransom4real (changes requested):
System tests might need updating and checks need to be added for the header

ransom4real (approved):

PR #443, opened 2022-09-01
Update `no match` page copy

ransom4real:
[Trello Card](https://trello.com/c/UviSxGM7)

### Context

The 'no match' page copy is wrong – it shouldn't be the "we could not find your TRN".

Update as per designs https://tra-digital-design-history.herokuapp.com/get-an-identity/no-match-journey/

### Guidance to review

- Ensure copy matches design
- Check test coverage

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased main
- [ ] Cleaned commit history


tvararu (commented):
This changes the content on the No match page for all users, not just the Identity ones.

Could we either fork the logic in the template itself, or if it's cleaner, create a separate template?

ransom4real:
> This changes the content on the No match page for all users, not just the Identity ones.
> 
> Could we either fork the logic in the template itself, or if it's cleaner, create a separate template?

Will create a separate template as well as include a test for this on the identity journey

tvararu (approved):

PR #446, opened 2022-09-01
Link the service name to the service URL

felixclack:
Currently, the service name is linked to the Find root path. This could
cause confusion for users who came to find their TRN from a different
service.

We accept a client_url in the params of the identity flow and can use
this, if present, to link the service name to the calling service
correctly.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -35,7 +35,7 @@ en:
         date_of_birth_form:
           attributes:
             date_of_birth:
-              blank: Enter your date of birth, formatted as a number
+              blank: Enter your date of birth
               born_after_1900: Enter a year of birth later than 1900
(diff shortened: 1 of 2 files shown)


tvararu (changes requested):
This isn't correct, as the `redirect_uri` is meant to be POSTed to and wouldn't work for this case.

James is adding an additional parameter that we'll be able to capture + use for this, we should update this PR afterwards. :+1:

tvararu:
New parameter is called `client_url`: https://github.com/DFE-Digital/get-an-identity/pull/89

rmdasilvaD2l (approved):

PR #447, opened 2022-09-01
Use Continue rather than Submit for the identity flow

felixclack:
When a user has arrived via the identity flow, the Submit button on the
check answers page is misleading, because the flow doesn't end there.

A more appropriate term is `Continue`.

This change ensures that when a TrnRequest is generated via the identity
flow, that we display the appropriate term.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


tvararu (approved):

tvararu:
Tests are breaking:

```
undefined local variable or method `when_i_press_the_submit_button'
```

PR #457, opened 2022-09-06
Add a Slack alert for Zendesk deletions

felixclack:
In a previous change, we scheduled the automatic deletion of old Zendesk
tickets to happen weekly.

We want to have some visibility of this process during the first few
weeks of it running to give us confidence that it is working as
expected.

A proposal is that a Slack alert 24 hours before the scheduled deletion
runs will give us sufficient notice to address any irregularities in the
expected count of tickets scheduled to be deleted.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/jobs/delete_zendesk_tickets_alert_job.rb
+++ b/app/jobs/delete_zendesk_tickets_alert_job.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+class DeleteZendeskTicketsAlertJob < ApplicationJob
+  include ActionView::Helpers::TextHelper
+  include Rails.application.routes.url_helpers
+
+  def perform
+    return unless FeatureFlag.active?(:slack_alerts)
+
+    count = ZendeskService.find_closed_tickets_from_6_months_ago.count
+    url = support_interface_zendesk_url(host: ENV.fetch("HOSTING_DOMAIN"))
+    message =
+      "There are #{pluralize(count, "Zendesk ticket")} scheduled for deletion tomorrow. #{url}"
+    SlackClient.create_message(message)
+  end
+end
--- a/config/schedule.yml
+++ b/config/schedule.yml
@@ -9,3 +9,7 @@ zendesk_health_alert:
 zendesk_delete:
   cron: "0 11 * * WED"
   class: "DeleteOldZendeskTicketsJob"
+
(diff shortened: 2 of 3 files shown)


app/jobs/delete_zendesk_tickets_alert_job.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+class DeleteZendeskTicketsAlertJob < ApplicationJob
+  include ActionView::Helpers::TextHelper
+  include Rails.application.routes.url_helpers
+
+  def perform
+    return unless FeatureFlag.active?(:slack_alerts)
+
+    count = ZendeskService.find_closed_tickets_from_6_months_ago.count
```
tvararu, line 9: This actually has a bug in it and it's affecting a few other things like the dashboard.

`.count` is the in-memory Ruby array count and as such is capped at 100 items, the pagination limit.

The request itself returns a `"count"` in the JSON body that contains the **real** count for the requested query; here it is for a date that returns 0 results so that we can actually see the JSON body:

```
irb(main):009:0>     date = (6.months.ago - 1.day).strftime("%Y-%m-%d")
=> "2022-03-05"
irb(main):010:0>     GDS_ZENDESK_CLIENT
irb(main):011:0>       .zendesk_client
irb(main):012:0>       .search(query: "updated<#{date} type:ticket status:closed")
irb(main):013:0>       .fetch
get https://teachingregulationagency.zendesk.com/api/v2/search?query=updated%3C2022-03-05+type%3Aticket+status%3Aclosed
  Accept: "application/json"
    Accept-Encoding: "gzip;q=1.0,deflate;q=0.6,identity;q=0.3"
    User-Agent: "ZendeskAPI Ruby 1.36.0"
  Status 200
    date: "Tue, 06 Sep 2022 10:31:53 GMT"
    ... headers ...
  cf-ray: "7466899b4d607587-LHR"
{"results"=>[], "facets"=>nil, "next_page"=>nil, "previous_page"=>nil, "count"=>0}
=> []                                                                  ^^^^^^^ here
```

Could you look into how we can surface that real count? And if you find a way, you can also update it on the dashboard / other places where it's used. :+1:

tvararu (approved):

PR #460, opened 2022-09-06
Build CSV export for Zendesk tickets

ransom4real:
[Trello Card](https://trello.com/c/PdVXjj1N)

Screenshot_2022-09-07_at_16 05 57

### Context

We need a way to export the deleted Zendesk tickets for reporting purposes.

- Export a CSV of deleted Zendesk tickets from `/support/zendesk`
- The CSV contains all the `ZendeskDeleteRequest`s
- The CSV contains 6 fields: ticket ID, group name, received at, closed at, enquiry type, no action required

### Guidance to review

Execute locally and attempt to export CSV

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/support_interface/zendesk_controller.rb
+++ b/app/controllers/support_interface/zendesk_controller.rb
@@ -1,3 +1,5 @@
+require "csv"
+
 module SupportInterface
   class ZendeskController < SupportInterfaceController
     include Pagy::Backend
@@ -33,6 +35,13 @@ module SupportInterface
         redirect_to support_interface_zendesk_path
       else
         render :confirm_deletion
+      end
+    end
+
+    def export
+      filename = "#{Time.zone.today}_deleted_zendesk_tickets.csv"
+      respond_to do |format|
+        format.csv { send_data ZendeskDeleteRequest.to_csv, filename: }
       end
     end
 
--- a/app/views/support_interface/zendesk/index.html.erb
+++ b/app/views/support_interface/zendesk/index.html.erb
@@ -83,6 +83,8 @@
     </span>
   </h2>
 
+  <p><%= govuk_link_to "Export to CSV", support_interface_zendesk_export_path(format: :csv) %></p>
+
   <%= govuk_pagination(pagy: @pagy) %>
 
   <%= govuk_table do |table|
--- a/app/models/zendesk_delete_request.rb
+++ b/app/models/zendesk_delete_request.rb
@@ -33,4 +33,21 @@ class ZendeskDeleteRequest < ApplicationRecord
     self.group_name = ticket.group.name
     self
   end
+
+  def self.to_csv
+    attributes = %w[
+      ticket_id
+      group_name
+      received_at
+      closed_at
+      enquiry_type
+      no_action_required
+    ]
+    CSV.generate(headers: true) do |csv|
+      csv << attributes.map(&:titleize)
+      all.find_each do |request|
+        csv << attributes.map { |attr| request.send(attr) }
+      end
+    end
+  end
 end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -30,6 +30,7 @@ Rails.application.routes.draw do
         :as => :validation_error
 
     get "/zendesk", to: "zendesk#index"
+    get "/zendesk/export", to: "zendesk#export"
     get "/zendesk/confirm-deletion", to: "zendesk#confirm_deletion"
     post "/zendesk/confirm-deletion", to: "zendesk#destroy"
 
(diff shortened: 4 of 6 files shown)


app/views/support_interface/zendesk/index.html.erb
```diff
@@ -81,6 +81,7 @@
     <span class="govuk-caption-m">
       Showing <%= @zendesk_delete_requests.size %> of <%= @total %> total
     </span>
+    <%= tag.a "Export to CSV", class: 'govuk-link', href: support_interface_zendesk_export_path(format: :csv) %>
```
tvararu, line 84: Does `govuk_link` work?

app/views/support_interface/zendesk/export.csv.erb
```diff
@@ -0,0 +1,5 @@
+<%- headers = ['Ticket ID', 'Group Name', 'Received At', 'Closed At', 'Enquiry Type', 'No Action Required'] -%>
+<%= CSV.generate_line headers %>
+<%- @zendesk_delete_requests.each do |request| -%>
+  <%= CSV.generate_line([request.ticket_id, request.group_name, request.received_at, request.closed_at, request.enquiry_type, request.no_action_required]) -%>
+<%- end -%>
```
tvararu, line 5: This file has no trailing newline; can you configure your editor to always add them?

(Sorry, I don't nit about this usually, but it looks like our linters aren't configured to catch this.)

app/controllers/support_interface/zendesk_controller.rb
```diff
@@ -36,6 +38,19 @@ def destroy
...
     end
 
+    def export
+      @zendesk_delete_requests = ZendeskDeleteRequest.all
+      respond_to do |format|
+        format.csv do
+          response.headers["Content-Type"] = "text/csv"
+          response.headers[
+            "Content-Disposition"
+          ] = "attachment; filename=zendesk_delete_requests.csv"
```
tvararu, line 48: `ZendeskDeleteRequest` is our internal name, but for our users, something like this would be more intuitive:

```suggestion
          ] = "attachment; filename=deleted_zendesk_tickets.csv"
```

tvararu (commented):
Could you add a system test for this to the existing Zendesk ones?

app/views/support_interface/zendesk/index.html.erb
ransom4real: Yes it does. Below is what the link looks like

Screenshot 2022-09-07 at 11 43 29

tvararu:
```suggestion
    <p>
      <%= govuk_link_to "Export to CSV", support_interface_zendesk_export_path(format: :csv) %>
    </p>
```

tvararu: Also this shouldn't in in the `h2`; could it be in a separate `<p>`?

app/controllers/support_interface/zendesk_controller.rb
tvararu: Maybe also prefix a timestamp with an ISO8166 date to the filename.

app/controllers/support_interface/zendesk_controller.rb
```diff
@@ -1,3 +1,6 @@
+require "csv"
+require "date"
```
tvararu, line 2: This shouldn't be needed.

```suggestion
```

app/controllers/support_interface/zendesk_controller.rb
```diff
@@ -36,6 +39,19 @@ def destroy
...
+      @zendesk_delete_requests = ZendeskDeleteRequest.all
+      respond_to do |format|
+        format.csv do
+          response.headers["Content-Type"] = "text/csv"
+          response.headers[
+            "Content-Disposition"
+          ] = "attachment; filename=#{Time.zone.now.iso8601}_zendesk_delete_tickets.csv"
+          render :export
+        end
+      end
```
tvararu, line 52:
```suggestion
      filename = "#{Date.today.iso8601}_deleted_zendesk_tickets.csv"
      respond_to do |format|
        format.csv do
          response.headers["Content-Type"] = "text/csv"
          response.headers[
            "Content-Disposition"
          ] = "attachment; filename=#{filename}"
          render :export
        end
      end
```

tvararu (changes requested):
Please add a system spec.

app/controllers/support_interface/zendesk_controller.rb
ransom4real: @tvararu as per your suggested, this will only return the date. Did we want a timestamp or just the date?

tvararu (approved):

PR #462, opened 2022-09-07
Use the count method provided by ZendeskAPI

felixclack:
The ZendeskAPI library includes a method, `count` on the
`ZendeskAPI::Collection` object.

We are not using this object but instead using `fetch` and working with
the JSON response from the API.

Rather than read the count value from the API response, it seems simpler
to lean on the interface provided by the library.

The change for this is small. In fact, most of the changes are in our
test code.

See [this comment](https://github.com/DFE-Digital/find-a-lost-trn/pull/457#discussion_r963545597) for the origin of this change.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


tvararu (approved):
Nice

PR #464, opened 2022-09-07
Healthcheck endpoint changed in statuscake

prabhakarreddy1234:
Statuscake now checks health of the app by pinging below urls.

https://find-a-lost-trn-production.london.cloudapps.digital/health/all
https://find-a-lost-trn.education.gov.uk/health/all

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/terraform/workspace_variables/production.tfvars.json
+++ b/terraform/workspace_variables/production.tfvars.json
@@ -9,9 +9,19 @@
   "postgres_database_service_plan": "small-ha-13",
   "hostnames": ["find-a-lost-trn"],
   "statuscake_alerts": {
-    "tra-flt-prod": {
+    "tra-flt-prod-1": {
       "website_name": "find-a-lost-trn-production",
-      "website_url": "https://find-a-lost-trn-production.london.cloudapps.digital/health",
+      "website_url": "https://find-a-lost-trn-production.london.cloudapps.digital/health/all",
+      "test_type": "HTTP",
+      "contact_group": [249142],
+      "check_rate": 30,
+      "trigger_rate": 0,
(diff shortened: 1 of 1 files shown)


prabhakarreddy1234:
[Check that `/health/all` returns a 200 during smoke test](https://trello.com/c/pU0GQ0fQ/678-check-that-health-all-returns-a-200-during-smoke-test)

terraform/workspace_variables/production.tfvars.json
```diff
@@ -9,9 +9,19 @@
   "postgres_database_service_plan": "small-ha-13",
   "hostnames": ["find-a-lost-trn"],
   "statuscake_alerts": {
-    "tra-flt-prod": {
+    "tra-flt-prod-1": {
       "website_name": "find-a-lost-trn-production",
-      "website_url": "https://find-a-lost-trn-production.london.cloudapps.digital/health",
+      "website_url": "https://find-a-lost-trn-production.london.cloudapps.digital/health/all",
```
tvararu, line 14: There should only be one check for the `find-a-lost-trn.education.gov.uk` domain; I don't think we need to test the cloudapps.digital one, as it's the same app, just a different CNAME.

tvararu (changes requested):

terraform/workspace_variables/production.tfvars.json
saliceti: It's interesting to keep both to identify if the problem is with paas or cloudfront (london.cloudapps.digital bypasses cloudfront)

tvararu: @saliceti ah ok, makes sense

tvararu (approved):

PR #465, opened 2022-09-07
Use ukprn for itt providers within trn searches

rmdasilvaD2l:
### Context

The ITT Provider name is currently used within TRN searches. If available the UKPRN should instead be used as names could change in the future.

Finds the UKPRN for selected ITT Providers and uses it within the find TRN search instead of the ITT Provider name

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


rmdasilvaD2l:
[Use UKPRN for ITT provider autocomplete](https://trello.com/c/zjxHUffl/633-use-ukprn-for-itt-provider-autocomplete)

PR #474, opened 2022-09-08
Add ADR for Support for Get an Identity

tvararu:
"Support for Get an Identity" is a new section that will live in Find's support. This is an ADR that documents some of our thinking for posterity.

### DIFF
--- a/adr/00010-support-for-get-an-identity.md
+++ b/adr/00010-support-for-get-an-identity.md
@@ -1,0 +1,49 @@
+# 10. Support for Get an Identity
+
+Date: 2022-09-08
+
+## Status
+
+Accepted
+
+## Context
+
+Get an Identity needs an interface to allow support users to link Get an
+Identity records to DQT records.
+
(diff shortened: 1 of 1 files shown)


adr/00010-support-for-get-an-identity.md
```diff
@@ -0,0 +1,49 @@
...
+The support section is "mostly stateless." The records exist in the Identity
+server and Find is just acting as a thin frontend to display and manage them in
+a user friendly way.
+
+### Consequences
+
+- Get an Identity will expose RESTful endpoints for user data
+- Find will integrate with Get an Identity's authentication system for
+  accessing and managing sensitive user data
+  - Staff/Admin user records in Find need to be matched/associated/created in
```
gunndabad, line 48: Identity will own support user accounts & this support area will need to integrate with ID using OIDC to authenticate those users.

aje54: any chance we can gets heads together on this please, to understand why identity support is being proposed to be in find.

PR #478, opened 2022-09-12
Allow importing ZendeskDeleteRequest by CSV

felixclack:
We have an export function on the ZendeskDeleteRequest object but no way
of importing them.

Recently, we ran into an issue where ZendeskDeleteRequests that were
incorrectly created in the pre-production environment need to be
migrated to the production environment.

I considered writing a data migration that would take the exported CSV
as an input and create the records in production.

It seemed to me to be a useful enough function of the support interface
to add it to the actual UI rather than doing it as a one-shot event.

I added a new form to the support interface that allows any staff member
to import a CSV of ZendeskDeleteRequests. It runs idempotently, so only
new records get created. Existing records aren't affected.

This seemed to be the most useful way of performing this migration.

As part of the design for this feature, I considered doing the actual
importing work in a background job. This would prevent the browser
timing out in the case of a large import.

At this stage, there isn't a reason to add the extra complexity. The
only export we need to import has 131 records and runs very fast in the
browser.

This could be revisited in the future if the file size gets too large.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ asdf plugin add redis
 asdf install
 ```
 
-If installing PostgreSQL via `asdf`, set up the `postgres` user:
+If installing PostgreSQL via `asdf`, you may need to set up the `postgres` user:
 
 ```bash
 pg_ctl start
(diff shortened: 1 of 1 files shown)


spec/system/zendesk_ticket_deletion_spec.rb
```diff
@@ -137,4 +147,23 @@ def and_i_should_be_able_to_export_tickets_to_csv
...
+      "ZendeskDeleteRequest CSV",
+      Rails.root.join("spec/support/zendesk_delete_requests.csv")
+    )
+    click_on("Upload")
+  end
+  alias_method :when_i_import_the_same_tickets_from_csv,
+               :when_i_import_tickets_from_csv
+
+  def then_i_should_see_the_imported_delete_request
+    expect(page).to have_content("Ticket 42")
```
rmdasilvaD2l, line 163: @felixclack I am wondering if for this test we should use another ticket number than 42. This ticket number is already used in the previous tests and should already be present on the page which may suggest that the upload is not working within the test

rmdasilvaD2l (changes requested):
Looks good, I have one query about one of the rspec tests

rmdasilvaD2l (approved):

PR #483, opened 2022-09-15
Identity support user page

malcolmbaig:
### Context

We're implementing a user show page in the support interface. This
page will eventually allow support staff to handle various aspects of
the user's identity. Initially we're going to add a skeleton page
with stubbed out data.

https://trello.com/c/w1q99XXi/694-identity-support-user-page

### Changes proposed in this pull request

- Add the govuk-summary-card component styles to match the styling of the prototype page (see image on Trello card)
- Stub out API response with test data for this page
- Implement the page itself, with non-functioning links in the summary card

### Guidance to review

- There are some potential refactors here, eg — putting the summary card into a view component and adding a model to wrap the JSON returned by the API. I've kept the implementation here fairly minimal on first pass.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [x] Tested by running locally

PR #484, opened 2022-09-16
Identity support user list page

rmdasilvaD2l:
### Context

A page for the Support section to list users stored within Get an Identity.

Adds a new page in the support section to list users stored within Get An Identity. Pagination is enabled for lists of over 100 users.

The Identity simulator has been moved to the `/support/identity/simulate` path.

Screenshot 2022-09-16 at 10 56 52

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/support_interface/identity_controller.rb
+++ b/app/controllers/support_interface/identity_controller.rb
@@ -27,7 +27,7 @@ module SupportInterface
     def callback
       flash[:success] = "You have completed a simulated Identity journey"
 
-      redirect_to support_interface_identity_path
+      redirect_to support_interface_identity_simulate_path
     end
 
     private
@@ -44,15 +44,15 @@ module SupportInterface
     end
 
     def redirect_url
-      support_interface_identity_callback_path
+      support_interface_identity_simulate_callback_path
     end
 
     def client_url
-      support_interface_identity_callback_path
+      support_interface_identity_simulate_callback_path
     end
 
     def previous_url
-      support_interface_identity_path
+      support_interface_identity_simulate_path
     end
   end
 end
--- a/app/controllers/support_interface/users_controller.rb
+++ b/app/controllers/support_interface/users_controller.rb
@@ -1,5 +1,14 @@
+# frozen_string_literal: true
 module SupportInterface
   class UsersController < SupportInterfaceController
+    include Pagy::Backend
+
+    def index
+      @all_users ||= IdentityApi.get_users
+      @total = @all_users.size
+      @pagy, @users = pagy_array(@all_users)
+    end
+
     def show
       @teacher = IdentityApi.get_teacher(uuid)
       @teacher_name =
--- a/app/helpers/support_interface_helper.rb
+++ b/app/helpers/support_interface_helper.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+module SupportInterfaceHelper
+  def user_info_link(user)
+    govuk_link_to(
+      user.full_name,
+      support_interface_identity_user_path(uuid: user.uuid),
+    )
+  end
+
+  def user_verification_status(user)
+    if user.name_verified?
+      govuk_tag(colour: "green", text: "VERIFIED")
+    else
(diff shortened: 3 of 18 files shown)


rmdasilvaD2l:
[Identity Support: User list page](https://trello.com/c/M7aT3xas/693-identity-support-user-list-page)

spec/system/identity_support_spec.rb
```diff
@@ -0,0 +1,168 @@
+require "rails_helper"
+
+RSpec.describe "Identity support", type: :system do
+  it "dislays a list of users", vcr: true, download: true do
```
malcolmbaig, line 4: typo: displays

app/controllers/support_interface/identity_controller.rb
```diff
@@ -27,7 +29,20 @@ def confirm
...
+      @total = @all_users.size
+      @pagy, @users = pagy_array(@all_users)
+    rescue IdentityApi::ApiError,
+           Faraday::ConnectionFailed,
+           Faraday::TimeoutError,
+           ActionController::ActionControllerError => e
+      Sentry.capture_exception(e)
+      # Do not redirect back to Get An Identity
+      render "errors/internal_server_error", status: :internal_server_error
     end
```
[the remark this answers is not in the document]
rmdasilvaD2l: @malcolmbaig yes good suggestion, I think this would make sense as the changes that you are already making include show page using a `UsersController`.

app/helpers/support_interface_helper.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module SupportInterfaceHelper
+  def user_info_link(user)
+    govuk_link_to(
+      [user["firstName"], user["lastName"]].join(" "),
```
tvararu, line 6: Should we create a wrapper class for the user object? That would allow us to only reference "firstName" in only one place, making it easier to deal with future API changes, and also would be a good place to place `def full_name` helpers and such.

app/controllers/support_interface/users_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+      @all_users ||= IdentityApi.get_users
+      @total = @all_users.size
+      @pagy, @users = pagy_array(@all_users)
+    rescue IdentityApi::ApiError,
+           Faraday::ConnectionFailed,
+           Faraday::TimeoutError,
+           ActionController::ActionControllerError => e
+      Sentry.capture_exception(e)
+      # Do not redirect back to Get An Identity
+      render "errors/internal_server_error", status: :internal_server_error
```
tvararu, line 16: Is this code doing anything? Without this, it'll just raise the error, it'll not get caught, Sentry will capture it and we'll display the 500 anyway. I guess one thing this is doing is saying "yeah, we expect this to fail, and we are being purposeful about the fact that we're displaying the 500 here for now instead of letting it just happen on its own"

spec/system/support_interface/users_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+  context "when there are more than 100 users" do
+    before do
+      user = {
+        "userId" => "37ee5357-fb84-478e-b750-bf552e5c8eed",
+        "email" => "kevin.e@example.com",
+        "firstName" => "Kevin",
+        "lastName" => "E",
+        "dateOfBirth" => "1990-01-01",
+        "trn" => nil
+      }
```
tvararu, line 19: Should we dry these? Most of these fields are repeated verbatim between all the different `user` variables in this file. It could be a hash of common attributes, `user_base`, that we `.merge(user_base)`.

app/helpers/support_interface_helper.rb
malcolmbaig: Agree with this — alluded to it briefly in my [show page PR](https://github.com/DFE-Digital/find-a-lost-trn/pull/483), but we could have something like an `app/models/user.rb` that accepts the JSON returned by the API in its constructor.

(Could live somewhere other than "model" if that's likely to be confusing at all).

tvararu (approved):

PR #495, opened 2022-09-20
Allow authenticating via Identity

felixclack:
We want to allow staff to authenticate via Identity to get access to the
support interface.

Identity has an OAuth2 server that we have a client for. Using Omniauth,
we are able to add Identity as an authentication provider.

As a first iteration, I opted to keep the data model simple and trust
the data returned by Identity. This means that we either match on email
for existing Staff records or create them when there is no match.

We have to override the redirect_uri parameter for the access token call.
OmniAuth assumes that the server will be able to handle a redirect_uri
containing querystring params. However, the implementation we are
working with needs them to be an exact match, so no querystring.

As a result, I redefined the method that retrieves the access token.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

Screenshot 2022-09-29 at 10 57 57 am

### DIFF
--- a/app/components/dqt_record_component.rb
+++ b/app/components/dqt_record_component.rb
@@ -5,19 +5,22 @@ class DqtRecordComponent < ViewComponent::Base
   end
 
   def full_name
-    @dqt_record.fetch("name", nil)
+    [
+      @dqt_record.fetch("firstName", nil),
+      @dqt_record.fetch("lastName", nil),
+    ].join(" ")
   end
 
   def date_of_birth
-    @dqt_record.fetch("dob", nil)&.to_date&.to_fs(:long_ordinal_uk)
+    @dqt_record.fetch("dateOfBirth", nil)&.to_date&.to_fs(:long_ordinal_uk)
   end
 
   def nino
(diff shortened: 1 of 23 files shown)


app/views/staff/shared/_links.html.erb
```diff
@@ -21,7 +21,9 @@
 
   <%- if devise_mapping.omniauthable? %>
     <%- resource_class.omniauth_providers.each do |provider| %>
-      <%= govuk_link_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider), method: :post %><br />
+      <% if FeatureFlag.active?(:identity_auth_service) %>
+        <%= govuk_button_to "Sign in with #{OmniAuth::Utils.camelize(provider)}", omniauth_authorize_path(resource_name, provider), method: :post %><br />
+      <% end %>
     <% end %>
   <% end %>
```
tvararu, line 28: Should the feature flag go around the whole `omniauthable?` section? I guess it doesn't make much difference.

tvararu (approved):
Looks pretty clean!

tvararu (approved):
Legend

PR #496, opened 2022-09-20
Update support user page to fetch data from Identity API

malcolmbaig:
### Context

The Identity API now has an endpoint for returning user details. We can make use of it on the user show page in support.

- Implement API interaction in the IdentityApi service class.
- Amend the system spec, adding a VCR recording of the interaction.

Screenshot_20220921_094502

### Guidance to review

n/a

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/support_interface/users_controller.rb
+++ b/app/controllers/support_interface/users_controller.rb
@@ -12,7 +12,7 @@ module SupportInterface
     def show
       @teacher = IdentityApi.get_teacher(uuid)
       @teacher_name =
-        "#{@teacher.fetch(:firstName)} #{@teacher.fetch(:lastName)}"
+        "#{@teacher.fetch("firstName")} #{@teacher.fetch("lastName")}"
     end
 
     private
--- a/app/controllers/support_interface/users_controller.rb
+++ b/app/controllers/support_interface/users_controller.rb
@@ -10,9 +10,7 @@ module SupportInterface
     end
 
     def show
-      @teacher = IdentityApi.get_teacher(uuid)
-      @teacher_name =
-        "#{@teacher.fetch("firstName")} #{@teacher.fetch("lastName")}"
+      @user = IdentityApi.get_teacher(uuid)
     end
 
     private
--- a/app/services/identity_api.rb
+++ b/app/services/identity_api.rb
@@ -5,15 +5,16 @@ class IdentityApi
 
   FIVE_SECONDS = 5
 
-  def self.get_teacher(_uuid)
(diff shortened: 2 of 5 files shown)


ransom4real:
@malcolmbaig you might no longer need this as i have these changes in my PR. Will send up for review shortly.

spec/system/support_interface/view_a_user_spec.rb
```diff
@@ -1,8 +1,15 @@
 # frozen_string_literal: true
 require "rails_helper"
 
-RSpec.describe "View a user in support", type: :system do
-  scenario "Staff user views user details" do
+RSpec.feature "User page in support", type: :system do
+  background do
+    @user_email = "kevin.e@digital.education.gov.uk"
+  end
```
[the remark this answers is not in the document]
malcolmbaig: I think that's a fair change. I try and avoid `let` in system specs, as their overuse can affect readability, but here the setup is minimal and needed before we even start writing steps in the scenario. Will amend 👍🏼

app/services/identity_api.rb
```diff
@@ -5,15 +5,16 @@ class ApiError < StandardError
...
+  def self.get_teacher(uuid)
+    if FeatureFlag.active?(:identity_api_always_timeout)
+      raise Faraday::TimeoutError, "Time-out feature flag enabled"
+    end
+
+    response = new.client.get("/api/v1/teachers/#{uuid}")
+
+    raise ApiError, response.reason_phrase unless response.status == 200
+
+    response.body
```
rmdasilvaD2l, line 17: @malcolmbaig we could use the result to populate the new User class in the same way as the `get_users` method? This way we have a mapping to the api response attributes in one place and do not need to reference them in the views

app/controllers/support_interface/users_controller.rb
```diff
@@ -12,7 +12,7 @@ def index
     def show
       @teacher = IdentityApi.get_teacher(uuid)
       @teacher_name =
-        "#{@teacher.fetch(:firstName)} #{@teacher.fetch(:lastName)}"
+        "#{@teacher.fetch("firstName")} #{@teacher.fetch("lastName")}"
```
rmdasilvaD2l, line 15: We could use the User class and the `full_name` method?

app/services/identity_api.rb
malcolmbaig: Thanks @rmdasilvaD2l , updated 👍🏼

PR #505, opened 2022-09-23
Match an Identity record to Kevin E

rmdasilvaD2l:
### Context

We want the Support for Get an Identity interface to allow users to match an Identity record to a DQT record. Before allowing users to specify a TRN, we can allow Kevin E's TRN to be matched for the user.

Allows a users to match to Kevin E's trn, skipping the steps to specify a trn, and sending the trn directly to the Identity API. Displays a success message and the DQT record of the user if successful.

https://user-images.githubusercontent.com/109349640/192007984-c3273f9a-7aa9-4158-beb5-ad9f15f12503.mov

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -42,6 +42,9 @@ Rails.application.routes.draw do
     get "/identity/simulate", to: "identity#new"
     get "/identity/users", to: "users#index"
     get "/identity/users/:uuid", to: "users#show", as: :identity_user
+    get "/identity/users/update/:uuid",
+        to: "users#update",
+        as: :identity_user_update
 
     get "/users/:uuid", to: "users#show", as: :user
 
--- a/app/components/dqt_record_component.html.erb
+++ b/app/components/dqt_record_component.html.erb
@@ -1,0 +1,39 @@
+<%
+  dqt_rows = []
+
+  name_row = {
+    key: { text: 'Official name' },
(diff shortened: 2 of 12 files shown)


rmdasilvaD2l:
[[Temporary] Identity Support: Instant matching to Kevin E](https://trello.com/c/z6IJY46h/131-temporary-identity-support-instant-matching-to-kevin-e)

config/routes.rb
```diff
@@ -42,6 +42,9 @@
     get "/identity/simulate", to: "identity#new"
     get "/identity/users", to: "users#index"
     get "/identity/users/:uuid", to: "users#show", as: :identity_user
+    get "/identity/users/update/:uuid",
+        to: "users#update",
+        as: :identity_user_update
```
tvararu, line 47: This PR is a temporary implementation, so this can probably go in, but I want to point out that this action shouldn't be a GET because it's not idempotent and causes a side effect

tvararu (approved):

config/routes.rb
rmdasilvaD2l: Thanks @tvararu , yes this was just temporary, the plan is to change it to a post request for https://trello.com/c/CPLxwXqk

PR #529, opened 2022-10-05
Create Zendesk tickets for Identity users

malcolmbaig:
### Context
For users going through the Identity auth flow, create a Zendesk ticket
if we don't find their TRN. This matches the existing Find a lost TRN
behaviour, with some Identity-specific changes to the ticket content.
https://trello.com/c/sptpIEil

- In the NoMatchController, if the trn_request is from get an identity,
  create a Zendesk ticket and a job to check the ticket for a TRN at a
  later time
- Add a separate template for Identity Zendesk tickets
- Add some specs to ensure the correct template is selected
- Include user-provided TRN in Identity Zendesk tickets

Relevant log output when simulating Identity journey in dev:
Screenshot_20221005_144329

### Guidance to review

- Can be tested locally in dev but until we've made the necessary updates to support different API keys for different sets of endpoints on Identity (https://trello.com/c/sptpIEil), you'll need to navigate directly to `/support/identity/simulate`. 
- You'll also need the fixed API key used by the `/api/find-trn/user/` endpoint (DM me if needed).

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/services/zendesk_service.rb
```diff
@@ -33,6 +33,41 @@ def self.destroy_tickets!(ids)
...
 
   def self.ticket_template(trn_request)
+    if trn_request.from_get_an_identity?
+      identity_ticket_template(trn_request)
+    else
+      find_a_lost_trn_ticket_template(trn_request)
+    end
+  end
+
+  def self.identity_ticket_template(trn_request)
```
steventux, line 43: Not necessarily for this PR but there's some template duplication here we could look to golf down in future.
It seems like the subject, custom_fields value and comment value differ, and in the case of the comment value we're only appending another line. 
Perhaps the templating could be DRYer with `Array#join("\n")`.
Not a blocker as this follows the previous implementation.

steventux (approved):

app/services/zendesk_service.rb
tvararu: Yeh, I think this second instance duplication is easier to understand than something more complex, but if there's a 3rd instance then we'll definitely revisit this

tvararu (approved):

PR #535, opened 2022-10-10
Remove error handling code in Identity controller

tvararu:
Rescuing and then capturing a sentry error + rendering a 500 is what happens by default. This code isn't very useful, because locally it actually causes a 500 page to be shown whereas you'd expect a detailed error page with a stacktrace and the like.

Review without whitespace.


malcolmbaig (approved):
👌🏼

PR #537, opened 2022-10-10
Check for expired access tokens on support pages

steventux:
### Context

With the identity auth service feature active we issue an access token when authenticating staff. This token has an expiry time of one hour, we should limit access to support pages based on this expiry.
https://trello.com/c/pwC7gEVw/229-staff-sign-in-flow-enhancements

Adds the `expires_at` value to the session as well as the access token.
Checks the `expires_at` value in the session against the current time when staff access support pages.

### Guidance to review

I couldn't find a solution to this which integrated well with existing Devise strategies, hence the `before_action` addition to `SupportInterfaceController`. Perhaps there's a neater way than this?

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/concerns/consumes_identity_users_api.rb
+++ b/app/controllers/concerns/consumes_identity_users_api.rb
@@ -9,7 +9,7 @@ module ConsumesIdentityUsersApi
 
   def identity_users_api_access_token
     if FeatureFlag.active?("identity_auth_service")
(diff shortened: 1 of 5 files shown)


steventux:
Thank you for review @felixclack I've amended PR in line with comments.

tvararu (approved):

PR #544, opened 2022-10-11
Remove the TRN query string parameter on the confirm TRN page

steventux:
### Context

The TRN is exposed in the URL as we redirect from successfully the updating the user to the DQT record lookup.

Store the TRN in session for use in the DQT lookup step.
Ensure we always remove from session

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


tvararu (approved):
The downside to storing it in the session is it's not robust if users open multiple tabs and attempt to work on multiple TRNs in parallel. For that probably storing it as a hidden field that is daisy-chained through the requests would work better (the hidden field would be resistant to tampering due to Rails' built-in stuff that prevents this)

This is better than the query string parameter though

steventux:
@tvararu Ah yes that's a good point, I did wonder about a hidden field but was a bit confused about how to hide the TRN value given we're doing a redirect to the next step. 
Would this mean we validate the TRN and then render (rather than redirect) a hidden field on `dqt_records#edit`?
Happy to amend to this.

tvararu:
@steventux yeah, we'd have to render instead of redirecting, but it's fine, we can stick to the current implementation for now :+1:

PR #562, opened 2022-10-18
Add preferred name to Get an Identity flow

malcolmbaig:
### Context

https://trello.com/c/ksJ8Sq3s outlines a need for users to be able to specify their preferred name in addition to their official and previous names.
Include relevant motivation and context.

Include a summary of the change.

### Guidance to review

Per commit recommended after you've glanced over the entire diff

Todo (separate PRs to follow):

- Validations in PreferredNameForm 
- Copy changes on official name page
- Double checking how we use official name in the codebase and changing to preferred name where appropriate (eg—emails).

### Checklist

- [x] Attach to Trello card
- [ ] Rebased main
- [x] Cleaned commit history


spec/components/trn_details_component_spec.rb
```diff
@@ -31,6 +33,10 @@
     expect(component.text).to include("Test User")
   end
 
+  it "renders preferred name" do
+    expect(component.text).to include("Ray Purchase")
```
steventux, line 37: Much better than star wars references.

spec/system/identity/user_matches_an_identity_spec.rb
```diff
@@ -0,0 +1,102 @@
+# frozen_string_literal: true
+require "rails_helper"
+
+RSpec.feature "Get an identity", type: :system do
+  include SummaryListHelpers
+
+  scenario "User matches an identity", vcr: { record: :once } do
```
steventux, line 7: Why do we use `vcr: { record: :once }` vs. `vcr: true` in the spec above?

steventux (approved):
:rocket: :shipit:

spec/system/identity/user_matches_an_identity_spec.rb
malcolmbaig: Yeah this should be more consistent really. I briefly entertained the idea that making the record mode explicit was easier to work with when needing to re-record a cassette (just a matter of changing the mode rather than writing out the entire hash). I'll write a PR which tidies this up and adds a command line switch for the record mode instead.

tvararu:
That commit log is chef's kiss! :man_cook: 

Excellent work @malcolmbaig!

PR #590, opened 2022-11-07
Allow changing names in Identity support

tvararu:
### Context

Implement Change name functionality in Identity support.

Add the controller, form model, a request spec.

We can't effectively system test this because of the way the short-lived OmniAuth keys work.

### Guidance to review

Prototype: https://find-a-lost-trn.herokuapp.com/support/101

image

image

image

### Checklist

- [x] Attach to Trello card https://trello.com/c/ipZ4SgJt/299-implement-change-name-functionality-in-identity-support
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/support_interface/change_name_controller.rb
+++ b/app/controllers/support_interface/change_name_controller.rb
@@ -1,0 +1,47 @@
+module SupportInterface
+  class ChangeNameController < SupportInterfaceController
+    include Pagy::Backend
+    include ConsumesIdentityUsersApi
+
+    layout "two_thirds"
+
+    def show
+      @user = identity_users_api.get_user(uuid)
+      @change_name_form = ChangeNameForm.new
+      @uuid = uuid
+    end
+
+    def update
+      @user = identity_users_api.get_user(uuid)
+      @change_name_form = ChangeNameForm.new(change_name_params)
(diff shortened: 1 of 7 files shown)


steventux (approved):
:rocket:

app/controllers/support_interface/change_name_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      @uuid = uuid
+    end
+
+    def update
+      @user = identity_users_api.get_user(uuid)
+      @change_name_form = ChangeNameForm.new(change_name_params)
+      @uuid = uuid
+      if @change_name_form.save
+        @user =
+          identity_users_api.update_user(
```
malcolmbaig, line 20: Would this be better placed in the body of the `save` method ?

malcolmbaig (approved):
Minor question left inline, otherwise LG 👍🏼

app/controllers/support_interface/change_name_controller.rb
tvararu: Hmm that's also a possibility, I think I like keeping request-y stuff in the controller; if it was a save to a local database I would have probably done it here

PR #620, opened 2022-11-22
Rework TRN matching

tvararu:
### Context

Users can currently find a matching TRN, then change their details, then continue to the end of the Identity journey with changed details. This is undesirable.

This moves the find_trn_using_api logic to the EnforceQuestionOrder concern and makes the relevant controllers always attempt to find a TRN if any of the user's data has changed after a submission.

It introduces a bug/undesirable flow where users can do this:

- Type in email
- Type in name
- Type in DOB
- Check answers
- Change email
- Now they get asked for their NI? What?
- Back button goes to DOB...

However, fixing that is quite involved and isn't part of this PR...

### Guidance to review

See commits; first two are not related. Review without whitespace.

You can probably ignore the .yml cassettes.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


steventux (approved):
:shipit: Looks sensible despite the slightly knotty bug mentioned in the PR description.

PR #652, opened 2022-12-14
Update the copy for the name question

felixclack:
Both the name and preferred name questions require an update to the
copy.

This makes the descriptions of what information to provide a little
clearer.

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history

### DIFF
--- a/app/views/preferred_name/edit.html.erb
+++ b/app/views/preferred_name/edit.html.erb
@@ -7,7 +7,7 @@
     <%= "Is #{@preferred_name_form.official_name} your preferred name?" %>
   </h1>
 
-  <p>You can tell us what you prefer to be called if it's different from the name on your offcial documents.</p>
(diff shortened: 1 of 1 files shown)


app/views/preferred_name/edit.html.erb
```diff
@@ -7,7 +7,9 @@
     <%= "Is #{@preferred_name_form.official_name} your preferred name?" %>
   </h1>
 
-  <p>We can use a different name in correspondence, and tell people what you prefer to be called.</p>
+  <p>You can tell us what you prefer to be called if it's different from the name on your offcial documents.</p>
```
tvararu, line 10:
```suggestion
  <p>You can tell us what you prefer to be called if it's different from the name on your official documents.</p>
```

tvararu (approved):

PR #1055, opened 2024-03-27
Clean up Makefile

malcolmbaig:
### Context
There's some old and/or broken stuff in the Makefile.

**Remove keyvault-related jobs from Makefile**

None of the keyvault printing / editing jobs work:
  
  - They invoke a configuration job read-keyvault-config that isn't
    defined.
  - Even if it was defined, the jobs would still fail because the Find
    app keyvault isn't structured like it is for other services. Specifically,
    it doesn't hold secrets in a serialized data structure sitting behind
    a single secret name defined in Azure.

As secrets can be added and updated via the Azure portal, simply remove
these jobs for now.

**Remove old PaaS jobs from Makefile**
These are jobs and scripts that were used when Find was still deployed
to GOVUK PaaS. They are no longer required now that Find deploys to AKS.

### Link to Trello card
https://trello.com/c/p5jfJKQF/1840-update-finds-key-vault-and-update-makefile

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [ ] Cleaned commit history


Makefile
```diff
@@ -80,50 +79,6 @@ install-fetch-config: ## Install the fetch-config script, for viewing/editing se
...
-	cf service ${POSTGRES_DATABASE_NAME} --guid
-	$(eval DB_INSTANCE_GUID=$(shell cf service ${POSTGRES_DATABASE_NAME} --guid))
-
-rename-postgres-service: read-deployment-config ## make dev rename-postgres-service NEW_NAME_SUFFIX=old CONFIRM_RENAME
-	$(if $(CONFIRM_RENAME), , $(error can only run with CONFIRM_RENAME))
-	$(if $(NEW_NAME_SUFFIX), , $(error NEW_NAME_SUFFIX is required))
-	cf target -s ${SPACE} > /dev/null
-	cf rename-service  ${POSTGRES_DATABASE_NAME} ${POSTGRES_DATABASE_NAME}-$(NEW_NAME_SUFFIX)
-
-restore-data-from-backup: read-deployment-config # make production restore-data-from-backup CONFIRM_RESTORE=YES BACKUP_FILENAME="find-a-lost-trn-production-pg-svc-2022-04-28-01"
```
malcolmbaig, line 120: Not 100% sure about this one - ok to delete?

saliceti: 100% yes
We will create a new solution for DR soon

saliceti (approved):

PR #600, opened 2022-11-09
Add Identity privacy notice 

tvararu:
### Context

This is the privacy page that we want to show while the user is in an Identity flow, specifically the Register for an NPQ one.

### Changes proposed in this pull request

See commits.

#### Usual notice

image

#### Identity flow

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history
- [x] Tested by running locally

PR #1305, opened 2026-07-22
Look up staff by uid and provider on sign-in

Lockyy:
### Context

Staff sign in through get-an-identity, but `Staff.from_identity` upserted the local record keyed on **email**. When a staff member changes their email in get-an-identity, the next sign-in no longer matches their existing record and a duplicate `staff` row is created. This mirrors the fix already shipped in sibling services ([CBL #600](https://github.com/DFE-Digital/check-childrens-barred-list/pull/600), [AYTQ #1126](https://github.com/DFE-Digital/access-your-teaching-qualifications/pull/1126)), which switched to keying on the identity provider's stable `uid`.

FLTRN needed more than a one-line swap: `staff` had no `uid`/`provider` columns and the `:identity` OmniAuth strategy never populated `uid`.

- **Add `uid` and `provider` to `staff`**, with a composite `unique` index on `[provider, uid]`. Nullable + unique is safe for the cutover: Postgres treats any row containing a NULL as distinct, so existing rows (both NULL) never collide, while backfilled rows enforce uniqueness.
- **Populate `uid` in the strategy** from the OIDC `sub` claim (`connect/userinfo`), verified against the get-an-identity auth-server source.
- **Rewrite `Staff.from_identity`** to look up by `[provider, uid]`, falling back to an *unclaimed* email row (both `uid` and `provider` still NULL) which it backfills, otherwise building a new record. This ordering never rebinds a `uid`/`provider` already attached to an email. Missing `uid` and a genuine email/uid conflict are raised as typed domain errors.
- **Normalise email at the boundary**: the strategy downcases `info.email`, and `from_identity` downcases again so lookups compare like-with-like even on paths (mocked auth) that bypass the strategy and never hit Devise's own downcasing.
- **Handle both error paths in the callback**: `IdentityEmailConflictError` and `IdentityMissingUidError` are reported to Sentry (capturing the two opaque `uid`s — **no email or other PII**) and the user is redirected to sign-in with a legible alert pointing at the support address.

**Assumptions / side effects**
- `uid` is populated lazily on each staff member's next sign-in via the email-fallback path, so no data backfill migration is required.
- The `email` unique index and Devise `:validatable` remain the backstop; a concurrent-login race on `save!` surfaces as a `RecordNotUnique` (fail-loud) rather than a duplicate.

### Guidance to review

- Start at `Staff.from_identity` — the three-branch lookup and the conflict check are the core.
- `spec/models/staff_spec.rb` covers create / return-by-uid+email-refresh / pre-cutover backfill / case-insensitive backfill / email-conflict / missing-uid. The two request specs cover the callback's Sentry capture (no PII) and graceful redirects.
- Worth a closer look: the safe-cutover ordering (never rebinding an email's existing uid) and the no-PII Sentry contexts.

### Link to Trello card

Tracked on the team project board: https://github.com/DFE-Digital/teaching-record-team-project-board/issues/351

### Checklist

- [x] Attach to Trello card
- [x] Rebased main
- [x] Cleaned commit history


app/models/staff.rb
```diff
@@ -53,15 +68,32 @@ class Staff < ApplicationRecord
...
-    staff = Staff.find_or_initialize_by(email: params.extra.raw_info.email)
+    uid = params.uid
+    provider = params.provider
+    email = params.info.email&.downcase
+    if uid.blank?
+      raise IdentityMissingUidError, "Identity auth response is missing a uid"
+    end
+
+    staff = find_by(provider:, uid:) ||
+            find_by(email:, uid: nil, provider: nil) ||
```
Lockyy, line 80: This line here allows existing records matching the provided email to be attached to the provided provider/uid combo, linking them going forwards to the new GAI record. 

This was necessary as we have an email uniqueness record and up until this point no records have been tied to uid/provider details. This step allows us to cut over from email look up to uid/provider lookup as users return.

gunndabad (approved):

Lockyy:
This ticket has been parked pending Entra ID work.

